PR #6, opened 2019-10-18
rm get started

lvuch:


yasker:
Need a `Signed-off-by`.

yasker:
Should we locate `Learn more` outside of the box? It's a bit weird to get in between the logos.

PR #11, opened 2019-10-21
Add Netlify config

lucperkins:
This PR adds an explicit TOML configuration for Netlify (so that the Netlify configuration is also in version control).


yasker:
Merged, thanks!

PR #18, opened 2019-12-06
Add tiered docs nav

lucperkins:
We can improve this over time but this is a first try at a tiered navigation system.

cc @yasker


yasker:
Merged, thanks!

Issue #17, opened 2019-12-06
Provide hierarchized documentation index


lucperkins:
Right now, the documentation index only has one "level." The nav and directory structure should be updated to accommodate multiple layers.

JasonvanBrackel:
@lucperkins I was actually about to put this ticket in for tracking.  You beat me to it

JasonvanBrackel:
@yasker, can you give me the ability to modify issues?  I want to assign myself to this.  Here's my first pass at this we can review at the next community meeting.  It encompasses all the current docs and will have us integrate the architecture documentation that's currently in the Rancher Blog.

- Longhorn
  - What is Longhorn?
  - FAQ
- Getting Started
  - Install Longhorn with Helm
- Install Longhorn
  - Install with Kubectl
  - Install with Helm
  - Customizing Longhorn Settings
- Longhorn Concepts
  - Volumes
  - Snapshots
  - Backups
    - Backup Stores
    - Backup Targets
    - Disaster Recover Volumes
- Longhorn Architecture
  - Longhorn UI
  - Longhorn Plugin
    - CSI
    - FlexVolume
    - Docker
  - Longhorn Manager
    - Volumes
    - Replicas
      - Replica Operations
      - Replica Rebuild
    - Controller (Engine)
      - Engine Controller
      - Volume Controller 
  - Backup and Restore
- Guides
  - Longhorn Operators Guide
    - Settings
    - Deploy Longhorn Standalone
      - iSCSI Support
    - Deploy Longhorn with Kubernetes
      - FlexVolume vs. CSI
    - Cloud Provider Notes
      - Google Kubernetes Engine
    - Longhorn CSI on RancherOS/CoreOS + RKE or K3s (Will reorg this some)
    - Multi-Disk Support
    - Storage Tags
      - Taints and Tolerations
    - Upgrades

  - Longhorn Users Guide
    - Workload Identification for a Volume
    - What happens when a Node Fails
    - Setup a Snapshot
    - Backup and Restore
      - Backup Stores
        - Setup an AWS Backup Store
        - Setup a NFS Backup Store
        - Setup a Minio Backup Store
      - Setup a Backup
      - Restore from a Backup
      - The restore-to-file command
    - Setup Disaster Recovery Volumes
    - Restoring Volumes for Kubernetes StatefulSets
    - Volume Operations

  - Troubleshooting Guide
    - Common Issues
    - UI
    - Manager and Engines
    - Drivers
      - CSI
      - FlexVolume

[...]

PR #31, opened 2020-03-20
Change directory structure to match documentation hierarchy

JasonvanBrackel:
Let's give this a go without all the extra changes and a properly squashed history.


yasker:
https://github.com/longhorn/website/issues/17

PR #36, opened 2020-04-03
Update docs to website so website can be the source of truth

JasonvanBrackel:


yasker:
Thanks @JasonvanBrackel 

Though it seems the content of the documents are not all updated.

For example, this PR still shows the requirement is `Kubernetes v1.8+` but we've already updated the requirement to `1.14+`. Also, the example of `kubectl -n longhorn-system get pod` wasn't updated too.

shuo-wu:
It seems that the modification in [this PR](https://github.com/longhorn/longhorn/pull/1155) is not covered.

JasonvanBrackel:
Ok, I see the issue now @yasker I sync'd everything that was in /docs.  And divided somethings to be consistent with the reorg, so we're good there.  I need to address the README.md file as well.  What other areas should I include in this PR?  Are there docs and other files in this and other repos that need to be considered?

JasonvanBrackel:
@yasker I'm going to go page by page through longhorn/longhorn and move anything over that should be on the site, like example scripts.

yasker:
@JasonvanBrackel We only need to deal with `docs` on `longhorn/longhorn`. If there are examples referenced by the docs, we can copy them over. Since we can change the examples in the manager's repo too, I would prefer to deal with them separately if possible.

PR #38, opened 2020-04-03
Convert JS to AlpineJS

lucperkins:
This PR removes jQuery from the site and replaces it with AlpineJS, which is much lighter.


yasker:
Hi @lucperkins , this PR has a conflict.

lucperkins:
Fixed

PR #48, opened 2020-04-14
Add longhorn/longhorn docs to longhorn.io/docs

catherineluse:
This PR has the following changes:

- Moves installation and upgrade information from longhorn/longhorn to longhorn.io/docs
- Removes passages about how to set up the FlexVolume storage driver (but keeps the docs about troubleshooting FlexVolume, and how to migrate from FlexVolume to CSI storage driver)
- Moves examples from longhorn/longhorn to longhorn.io/docs
- Creates a contributing page with the source code, invitation to collaborate, and license information
- Replaced the example of a successful CSI-based deployment after a Helm installation of Longhorn, because the one on longhorn/longorn was more recently updated


yasker:
@catherineluse Build failed due to dead links.

yasker:
https://github.com/longhorn/longhorn/issues/1131

PR #58, opened 2020-04-16
Add doc default-disk-and-node-config.md

shuo-wu:
https://github.com/longhorn/longhorn/issues/1168

### DIFF
--- a/content/docs/0.8.0/users-guide/default-disk-and-node-config.md
+++ b/content/docs/0.8.0/users-guide/default-disk-and-node-config.md
@@ -1,0 +1,41 @@
+# Default disks and node configuration
+
+## Overview
+
+This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes **using Kubernetes label and annotation**, instead of using Longhorn API or UI.
+
+This feature is available since v0.8.1.
+
+## Usage
+
+### Add node tags for new nodes
+
+1. Scale up the Kubernetes cluster. Then the newly added nodes contain no node tag.
+2. Add the annotations to the new Kubernetes nodes. 
+    - The annotation format is `node.longhorn.io/default-node-tags: <node tag list with JSON string format>`. For example:
+    ```
+    node.longhorn.io/default-node-tags: '["fast","storage"]'
+    ``` 
+3. Wait for Longhorn syncing the node tag automatically.
+
+### Customize default disks for new nodes
+
+1. Enable the Longhorn setting `Create Default Disk on Labeled Nodes` to enable the feature. After the feature is enabled, the default disk will not be automatically created on the new nodes unless a certain label was set. 
+2. Add new nodes to the Kubernetes cluster, e.g. by using Rancher or Terraform, etc.
+3. Patch the label and annotation to each new node to define the default disks.
+    - Longhorn relies on the label to decide how to customize default disks. Since we choose using annotation here, the label should be:
+    ```
+    node.longhorn.io/create-default-disk: 'config'
+    ```
+    - The annotation is used to specify the configuration of default disks. The format is `node.longhorn.io/default-disks-config: <disks configuration with JSON string format>`. For example:
+    ```
+    node.longhorn.io/default-disks-config: 
+    '[{"path":"/mnt/disk1","allowScheduling":false},
+      {"path":"/mnt/disk2","allowScheduling":false,"storageReserved":1024,"tags":["ssd","fast"]}]'
+    ```
+4. Wait for Longhorn creating the customized default disks automatically.
+
+### Notice
+1. This feature will only take effect when there is no disks/tags on the node.
+2. Longhorn will not keep the node label/annotations in sync with the current Longhorn node disks/tags.
+
--- a/content/docs/0.8.0/users-guide/default-disk-and-node-config.md
+++ b/content/docs/0.8.0/users-guide/default-disk-and-node-config.md
@@ -1,41 +1,79 @@
-# Default disks and node configuration
+---
+title: Configuring Defaults for Nodes and Disks
+weight: 1
+---
 
-## Overview
+_Available as of v0.8.1_
 
-This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes **using Kubernetes label and annotation**, instead of using Longhorn API or UI.
+This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes using Kubernetes labels and annotations instead of the Longhorn API or UI.
 
-This feature is available since v0.8.1.
+Customizing the default configurations for disks and nodes is useful for scaling the cluster because it eliminates the need to configure Longhorn manually for each new node if the node contains more than one disk, or if the disk configuration is different for new nodes.
 
-## Usage
+Longhorn will not keep the node labels or annotations in sync with the current Longhorn node disks or tags. Nor will Longhorn keep the node disks or tags in sync with the nodes, labels or annotations after the default disks or tags have been created.
 
-### Add node tags for new nodes
+### Adding Node Tags to New Nodes
 
-1. Scale up the Kubernetes cluster. Then the newly added nodes contain no node tag.
-2. Add the annotations to the new Kubernetes nodes. 
-    - The annotation format is `node.longhorn.io/default-node-tags: <node tag list with JSON string format>`. For example:
+When a node does not have a tag, you can use a node annotation to set the node tags, as an alternative to using the Longhorn UI or API.
+
+1. Scale up the Kubernetes cluster. The newly added nodes contain no node tags.
+2. Add annotations to the new Kubernetes nodes that specify what the default node tags should be. The annotation format is:
+
+    ```
+    node.longhorn.io/default-node-tags: <node tag list with JSON string format>
+    ```
+    For example:
+    
     ```
     node.longhorn.io/default-node-tags: '["fast","storage"]'
     ``` 
-3. Wait for Longhorn syncing the node tag automatically.
+3. Wait for Longhorn to sync the node tag automatically.
 
-### Customize default disks for new nodes
+> **Result:** If the node tag list was originally empty, Longhorn updates the node with the tag list, and you will see the tags for that node updated according to the annotation. If the node already had tags, you will see no change to the tag list.
 
-1. Enable the Longhorn setting `Create Default Disk on Labeled Nodes` to enable the feature. After the feature is enabled, the default disk will not be automatically created on the new nodes unless a certain label was set. 
-2. Add new nodes to the Kubernetes cluster, e.g. by using Rancher or Terraform, etc.
-3. Patch the label and annotation to each new node to define the default disks.
-    - Longhorn relies on the label to decide how to customize default disks. Since we choose using annotation here, the label should be:
+### Customizing Default Disks for New Nodes
+
+Longhorn uses the **Create Default Disk on Labeled Nodes** setting to enable default disk customization.
+
+If the setting is disabled, Longhorn will create a default disk using `setting.default-data-path` on all new nodes.
+
+If the setting is enabled, Longhorn will decide to create the default disks or not, depending on the node's label value of `node.longhorn.io/create-default-disk`.
+    
+- If the node's label value is `true`, Longhorn will create the default disk using `settings.default-data-path` on the node. If the node already has existing disks, Longhorn will not change anything.
+- If the node's label value is `config`, Longhorn will check for the `node.longhorn.io/default-disks-config` annotation and create default disks according to it. If there is no annotation, or if the annotation is invalid, or the label value is invalid, Longhorn will not change anything.
+
+The value of the label will be in effect only when the setting is enabled.
+
+If the `create-default-disk` label is not set, the default disk will not be automatically created on the new nodes when the setting is enabled.
+
+The configuration described in the annotation only takes effect when there are no existing disks or tags on the node.
+
+If the label or annotation fails validation, the whole annotation is ignored. 
+
+> **Prerequisite:** The Longhorn setting **Create Default Disk on Labeled Nodes** must be enabled.
+
+1. Add new nodes to the Kubernetes cluster.
+2. Add the label to the node. Longhorn relies on the label to decide how to customize default disks:
+
     ```
     node.longhorn.io/create-default-disk: 'config'
     ```
-    - The annotation is used to specify the configuration of default disks. The format is `node.longhorn.io/default-disks-config: <disks configuration with JSON string format>`. For example:
+    
+3. Then add an annotation to the node. The annotation is used to specify the configuration of default disks. The format is:
+    
+    ```
+    node.longhorn.io/default-disks-config: <disks configuration with JSON string format>
+    ```
+    
+    For example, the following disk configuration can be specified in the annotation:
+    
     ```
     node.longhorn.io/default-disks-config: 
-    '[{"path":"/mnt/disk1","allowScheduling":false},
-      {"path":"/mnt/disk2","allowScheduling":false,"storageReserved":1024,"tags":["ssd","fast"]}]'
+    '[{"path":"/mnt/disk1","allowScheduling":true},
+    {"name":"fast-ssd-disk", "path":"/mnt/disk2","allowScheduling":false,"storageReserved":10485760,"tags":["ssd","fast"]}]'
     ```
-4. Wait for Longhorn creating the customized default disks automatically.
 
-### Notice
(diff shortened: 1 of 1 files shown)


shuo-wu:
Wait for the directory `docs/0.8.1` creation.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,42 @@
...
+    ```
+    - The annotation is used to specify the configuration of default disks. The format is `node.longhorn.io/default-disks-config: <disks configuration with JSON string format>`. For example:
+    ```
+    node.longhorn.io/default-disks-config: 
+    '[{"path":"/mnt/disk1","allowScheduling":false},
+      {"path":"/mnt/disk2","allowScheduling":false,"storageReserved":1024,"tags":["ssd","fast"]}]'
+    ```
+4. Wait for Longhorn creating the customized default disks automatically.
+
+## History
```
yasker, line 37: I don't think we need a reference to the issues here.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,42 @@
+# Default disks and node configuration
+
+## Overview
+
+This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes **using Kubernetes label and annotation**, instead of using Longhorn API or UI.
+For more details, see [this enhancement doc](https://github.com/longhorn/longhorn/blob/v0.8.1/enhancements/20200319-default-disks-and-node-configuration.md).
```
yasker, line 6: Not sure if we should refer to the design doc in the formal documentation.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,42 @@
...
+2. Add the annotations to the new Kubernetes nodes. 
+    - The annotation format is `node.longhorn.io/default-node-tags: <node tag list with JSON string format>`. For example:
+    ```
+    node.longhorn.io/default-node-tags: '["fast","storage"]'
+    ``` 
+3. Wait for Longhorn syncing the node tag automatically.
+
+### Customize default disks for new nodes
+
+1. Enable the Longhorn setting `Create Default Disk on Labeled Nodes` to prevent the default disk to be created automatically on new nodes.
```
yasker, line 22: `to enable the feature. After the feature is enabled, the default disk will not be automatically created on the new nodes unless a certain label was set.`

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,42 @@
...
+    ```
+    node.longhorn.io/default-node-tags: '["fast","storage"]'
+    ``` 
+3. Wait for Longhorn syncing the node tag automatically.
+
+### Customize default disks for new nodes
+
+1. Enable the Longhorn setting `Create Default Disk on Labeled Nodes` to prevent the default disk to be created automatically on new nodes. 
+2. Add new nodes to the Kubernetes cluster, e.g. by using Rancher or Terraform, etc.
+3. Patch the label and annotation to each new node to define the default disks.
```
yasker, line 24: mentioned that the feature will only take effect there is no disks on the node.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
shuo-wu: Can I say `If you are interested in the feature design, see here for details`

yasker: The problem is we're not going to update the design doc, but we will always keep doc up to date after we add more features etc. So I prefer to not have a permanent link here to the design doc.

Of course, the design doc can help contributors to understand the feature better. But if the user asks for more details, we can also point them to the design doc. But that should be very rare if our official document serves its purpose well.

shuo-wu: OK. Then I will remove this link.

catherineluse:
I edited the doc for grammar/style and added a few more details from the design doc. It looks good to me but should probably be re-reviewed to make sure I didn't introduce any errors.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
+3. Wait for Longhorn to sync the node tag automatically.
+
+> **Result:** If the node tag list was originally empty, Longhorn updates the node with the tag list, and you will see the tags for that node updated according to the annotation. If the node already had tags, you will see no change to the tag list.
+
+### Customizing Default Disks for New Nodes
+
+Longhorn uses the **Create Default Disk on Labeled Nodes** setting to enable default disk customization.
+
+- If the setting is enabled, Longhorn waits for the default disk customization to be set.
+- If the setting is disabled, Longhorn creates the default disk for new nodes that don't have existing disks. Then Longhorn relies on the value of the node label node.longhorn.io/create-default-disk to decide how to customize the default disks.
```
yasker, line 38: If the setting is disabled, Longhorn will create the default disk for any newly added nodes, using the disk path specified by settings `Default Data Path`.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
+
+If the value of the label is `config`, the annotation will be parsed and used as the default disk customization. If the value is the boolean value `true`, the data path setting will be used for the default disk. Other values will be treated as false and no default disk will be created.
```
yasker, line 40: The value and labels will be in effect only when the setting is enabled.

Also, we can have a section about what will happen if the user sets the label to `true` instead of `config`.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
+    
+    ```
+    node.longhorn.io/default-disks-config: <disks configuration with JSON string format>
+    ```
+    
+    For example, the following disk configuration can be specified in the annotation:
+    
     ```
     node.longhorn.io/default-disks-config: 
     '[{"path":"/mnt/disk1","allowScheduling":false},
```
yasker, line 67: Let use this example instead:

```
node.longhorn.io/default-disks-config: 
'[{"path":"/mnt/disk1","allowScheduling":true},
  {"name":"fast-ssd-disk", "path":"/mnt/disk2","allowScheduling":false,"storageReserved":10485760,"tags":["ssd","fast"]}]'
```

We've added the ability to add name to disk recently. If user specify the same name for different disks, the configuration will be treated as invalid.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
+The default disk will not be automatically created on the new nodes unless the `create-default-disk` label is set.
 
-### Customize default disks for new nodes
+The configuration described in the annotation only takes effect when there are no existing disks or tags on the node.
+
+If the label or annotation fails validation, the whole annotation is ignored. 
+
+> **Prerequisite:** The Longhorn setting **Create Default Disk on Labeled Nodes** must be enabled.
+
+1. Add new nodes to the Kubernetes cluster, for example, by using Rancher or Terraform.
```
yasker, line 50: Don't need the `for example` part.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
yasker: Also, the node.longhorn.io/create-default-disk needs to be quoted `node.longhorn.io/create-default-disk`.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
+_Available as of v0.8.1_
 
-This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes **using Kubernetes label and annotation**, instead of using Longhorn API or UI.
+This feature allows the user to customize the default disks and node configurations in Longhorn for newly added nodes using Kubernetes labels and annotations instead of the Longhorn API or UI.
 
-This feature is available since v0.8.1.
+Customizing the default configuations for disks and nodes is useful for scaling the cluster because it eliminates the need to configure Longhorn manually for each new node if the node contains more than one disk, or if the disk configuration is different for new nodes.
 
-## Usage
+> **Note:** Longhorn will not keep the node labels or annotations in sync with the current Longhorn node disks or tags.
```
yasker, line 12: Add something like: `Nor will Longhorn keep the node disks or tags in sync with the nodes labels or annotations after the default disks or tags have been created`

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -1,41 +1,78 @@
...
 
-### Add node tags for new nodes
+### Adding Node Tags to New Nodes
 
-1. Scale up the Kubernetes cluster. Then the newly added nodes contain no node tag.
-2. Add the annotations to the new Kubernetes nodes. 
-    - The annotation format is `node.longhorn.io/default-node-tags: <node tag list with JSON string format>`. For example:
+When a node does not have a tag, Longhorn uses a node annotation to set the node tags.
```
yasker, line 16: `user can use a node annotation to set the node tags, as an alternative to using the Longhorn UI/API`. Since the user can also use Longhorn UI to set it.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: updated

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: changed to:

> When a node does not have a tag, you can use a node annotation to set the node tags, as an alternative to using the Longhorn UI or API.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: updated

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: changed it to:

> The value and labels will be in effect only when the setting is enabled.
> 
> If the value of the label is `config`, the annotation will be parsed and used as the default disk customization. 
> 
> If the label is set to `true`, Longhorn will use `settings.default-data-path` to create a default disk if the node has no disks.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: > 1. Add new nodes to the Kubernetes cluster. Rancher or Terraform can be used to add nodes.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: For example, the following disk configuration can be specified in the annotation:
    
    ```
    node.longhorn.io/default-disks-config: 
    '[{"path":"/mnt/disk1","allowScheduling":true},
    {"name":"fast-ssd-disk", "path":"/mnt/disk2","allowScheduling":false,"storageReserved":10485760,"tags":["ssd","fast"]}]'
    ```

    > **Note:** If the same name is specified for different disks, the configuration will be treated as invalid.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
yasker: I meant, just `Add new nodes to the Kubernetes cluster.` would be enough.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,81 @@
...
+3. Wait for Longhorn to sync the node tag automatically.
+
+> **Result:** If the node tag list was originally empty, Longhorn updates the node with the tag list, and you will see the tags for that node updated according to the annotation. If the node already had tags, you will see no change to the tag list.
+
+### Customizing Default Disks for New Nodes
+
+Longhorn uses the **Create Default Disk on Labeled Nodes** setting to enable default disk customization.
+
+- If the setting is enabled, Longhorn waits for the default disk customization to be set.
+- If the setting is disabled, Longhorn will create the default disk for any newly added nodes, using the disk path specified by settings `Default Data Path`. Then Longhorn relies on the value of the node label `node.longhorn.io/create-default-disk` to decide how to customize the default disks.
```
yasker, line 38:
```
- If the setting is disabled, Longhorn will create the default disk for any newly added nodes, using the disk path specified by settings `Default Data Path`.
```
If the setting is disabled. No label or annotation will be considered as configuration.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,81 @@
...
+
+The value and labels will be in effect only when the setting is enabled.
```
yasker, line 40: `The value of the label`

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
```diff
@@ -0,0 +1,81 @@
...
+
+If the value of the label is `config`, the annotation will be parsed and used as the default disk customization. 
+
+If the label is set to `true`, Longhorn will use `settings.default-data-path` to create a default disk if the node has no disks.
+
+Other values will be treated as false and no default disk will be created.
+
+The default disk will not be automatically created on the new nodes unless the `create-default-disk` label is set.
```
yasker, line 48: This is not true if the setting was disabled.

Basically it's like this:
1. If the setting is disabled, Longhorn will create default disk using setting.default-data-path on all new nodes.
2. If the setting is enabled, Longhorn will decide creating the default disks or not, depends on the node's label value
    1. If the node's label value is `true`, Longhorn will create default disk using settings.default-data-path on the node.
    2. If the node's label value is `config`, Longhorn will check for the annotation and create default disks according to it
        1. if there is no annotation or annotation is invalid or the label value is invalid, Longhorn will not change anything.

yasker:
@catherineluse We can just have two commits in this PR, one is from Shuo's original one, one is yours.

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: updated

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: added sentence

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: updated

content/docs/0.8.0/users-guide/default-disk-and-node-config.md
catherineluse: ### Customizing Default Disks for New Nodes

Longhorn uses the **Create Default Disk on Labeled Nodes** setting to enable default disk customization.

If the setting is disabled, Longhorn will create a default disk using `setting.default-data-path` on all new nodes.

If the setting is enabled, Longhorn will decide to create the default disks or not, depending on the node's label value of `node.longhorn.io/create-default-disk`.
    
- If the node's label value is `true`, Longhorn will create the default disk using `settings.default-data-path` on the node. If the node already has existing disks, Longhorn will not change anything.
- If the node's label value is `config`, Longhorn will check for the `node.longhorn.io/default-disks-config` annotation and create default disks according to it. If there is no annotation, or if the annotation is invalid, or the label value is invalid, Longhorn will not change anything.

The value of the label will be in effect only when the setting is enabled.

If the `create-default-disk` label is not set, the default disk will not be automatically created on the new nodes.

The configuration described in the annotation only takes effect when there are no existing disks or tags on the node.

If the label or annotation fails validation, the whole annotation is ignored.

yasker:
```
If the create-default-disk label is not set, the default disk will not be automatically created on the new nodes when the setting is enabled.
```

catherineluse: updated

PR #59, opened 2020-04-16
Add Algolia search bar

lucperkins:
cc @yasker 

They actually finished faster than I expected. Yay!


yasker:
Great!

Though I found the search bar is unclickable on my phone. It will just disappear when I click it.

lucperkins:
@yasker I'm not sure there's a great solution to that. In fact, I think I may hide it on mobile for now.

yasker:
@lucperkins that will do for now. Thanks.

PR #80, opened 2020-04-24
Fix EKS ingress instructions

meldafrawi:


catherineluse:
I see this error in the logs:

```
target does not exist --- docs/0.8.0/install/viewing-ui/index.html --> ../users-guide#longhorn-ingress
```

Can you please change the link like this:
```
See more about how to create an Nginx ingress controller with basic authentication [here](../../users-guide/longhorn-ingress)
```

And squash the commits into one commit?

meldafrawi:
@catherineluse fixed, thank you.

catherineluse:
LGTM

PR #86, opened 2020-04-30
Re-add missing diagram to architecture doc

catherineluse:


yasker:
Is it possible to replace `Docker` with `Kubernetes` in the diagram?

catherineluse:
Lauren said she doesn't have the source file, but she said she would redraw it and change Docker volume to Kubernetes volume.

catherineluse:
I have added Lauren's updated diagram to the architecture page and to Longhorn.io

yasker:
Sorry, I just realized I missed one thing. the rightmost engine is pointing to two replicas on the same node, so does the left-most engine

catherineluse:
I updated it with Lauren's most recent version

yasker:
LGTM. Need to resolve conflict.

catherineluse:
it's fixed now

PR #88, opened 2020-05-01
Add explanation on how to use a self signed ssl certificate for s3

joshimoo:
Our default minio-test backupstore is setup to use such a certificate
and can be used as an example.

Signed-off-by: Joshua Moody <joshua.moody@rancher.com>

### DIFF
--- a/content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
+++ b/content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
@@ -102,7 +102,8 @@ We provides two testing purpose backupstore based on NFS server and Minio S3 ser
     data:
       AWS_ACCESS_KEY_ID: bG9uZ2hvcm4tdGVzdC1hY2Nlc3Mta2V5 # longhorn-test-access-key
       AWS_SECRET_ACCESS_KEY: bG9uZ2hvcm4tdGVzdC1zZWNyZXQta2V5 # longhorn-test-secret-key
-      AWS_ENDPOINTS: aHR0cDovL21pbmlvLXNlcnZpY2UuZGVmYXVsdDo5MDAw # http://minio-service.default:9000
+      AWS_ENDPOINTS: aHR0cHM6Ly9taW5pby1zZXJ2aWNlLmRlZmF1bHQ6OTAwMA== # https://minio-service.default:9000
+      AWS_CERT: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUREekNDQWZlZ0F3SUJBZ0lSQU91d1oybTZ6SXl3c1h2a2UyNS9LYzB3RFFZSktvWklodmNOQVFFTEJRQXcKRWpFUU1BNEdBMVVFQ2hNSFFXTnRaU0JEYnpBZUZ3MHlNREEwTWpVd01qRTJNalphRncweU1UQTBNalV3TWpFMgpNalphTUJJeEVEQU9CZ05WQkFvVEIwRmpiV1VnUTI4d2dnRWlNQTBHQ1NxR1NJYjNEUUVCQVFVQUE0SUJEd0F3CmdnRUtBb0lCQVFEWkpyWVUraVJhc1huSExvb1d0Tm9OQWpxN0U3YWNlQTJQZnQ1ZFM3aExzVUtCbExMOVVQMmUKZ0QrVFl3RmtCWVJNU3BsV0tNT0tuWEErNDZSVkRwSkhwSTF4YjhHNDV0L3gzVXhVaWc2WUFVbDBnTFV6N01rMQpYSUtRaWZZUllaL0FjUzJqU0VOYjRISFJ1aFZ5NzV0ZDdCaXNhd2J2TTJwTXI0dWNSR1lwZ3J6Z2V2eFBXSHZ1CnkxT29yRnIvNjFwV28wcG9aSXhtRmM2YXMzekw0NWlrRzRHN1A2ejJPamc4NGdrdnR4RFUzYVdmWXRNb3VhL3gKQVhkRlRCd2NqMkNHMHJtdmd4cE5KeEx5Kzl5NDVLVGU1SFlSd0xxUjVCeWtnVGt2RGplcWdXTnJyQWdCL3lLTApwU1ZjRmZkKzBWNjhyQmtNMEt3VlQ3bXF2WWRsZDVrTkFnTUJBQUdqWURCZU1BNEdBMVVkRHdFQi93UUVBd0lDCnBEQVRCZ05WSFNVRUREQUtCZ2dyQmdFRkJRY0RBVEFQQmdOVkhSTUJBZjhFQlRBREFRSC9NQ1lHQTFVZEVRUWYKTUIyQ0ZXMXBibWx2TFhObGNuWnBZMlV1WkdWbVlYVnNkSWNFZndBQUFUQU5CZ2txaGtpRzl3MEJBUXNGQUFPQwpBUUVBdDBQYjM5QWliS0EyU1BPTHJHSmVRVlNaVTdZbFUyU0h0M2lhcFVBS1Rtb2o1RTQrTU8yV2NnbktoRktrCnNxeW9CYjBPYTNPMHdXUnRvVnhNUGdPQS9FaXNtZExQWmJIVTUzS2w3SDVWMm8rb0tQY0YydTk2ajdlcUZxSkUKMlltQkpBTHlUVks5LzZhS1hOSnRYZE5xRmpPMWJRcDJRTURXbjQyZGgyNjJLZmgvekM4enRyK0h4RzhuTVpQQwpsZUpxbzU3S0tkZ0YvZHVBWTdUaUI2cThTelE4RmViMklQQ2FhdVVLNzdBZ0d5b3kzK1JuWkdZV2U1MG1KVnN6CmdkQTFURmg0TVdMeUxWSFdIbnl2cEFvTjJIUjQrdzhYRkpJS2VRRFM1YklJM1pFeU5OQUZNRDg0bTVReGY4cjUKMEovQWhXTVVyMFUwcCtiRi9KM3FDQVNSK3c9PQotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
     ```
     For more information on creating a secret, see [the Kubernetes documentation.](https://kubernetes.io/docs/concepts/configuration/secret/#creating-a-secret-manually) The secret must be created in the `longhorn-system` namespace for Longhorn to access it.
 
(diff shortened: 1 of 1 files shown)


yasker:
@catherineluse It seems `content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md` and `content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md` has duplicate content.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md
```diff
@@ -108,6 +108,13 @@ Please follow [the Kubernetes document](https://kubernetes.io/docs/concepts/conf
 
 Notice the secret must be created in the `longhorn-system` namespace for Longhorn to access.
 
+### Using a self signed SSL certificate for S3 communication
+If you want to use a self signed ssl certificate you can expose an AWS_CERT environment variable that contains your public certificate to the longhorn pods.
+we recommend you do so by adding the AWS_CERT key/val to the `minio-secret` or `aws-secret` and making it available to the `longhorn-system` namespace.
```
yasker, line 113: `We`. `key/value pair`

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md
```diff
@@ -108,6 +108,13 @@ Please follow [the Kubernetes document](https://kubernetes.io/docs/concepts/conf
...
+For an example of using a self signed ssl certificate, have a look at our `minio-backupstore.yaml` linked above.
```
yasker, line 114: YAML file wasn't updated.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md
```diff
@@ -108,6 +108,13 @@ Please follow [the Kubernetes document](https://kubernetes.io/docs/concepts/conf
...
+It's important to note that the Certificate needs to be in PEM format and must either be a certificate authority,
+or the certificate chain including the certificate authority must be included by concatenating all relevant certificates.
```
yasker, line 116: Grammar.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md
joshimoo: I added the AWS_CERT under the minio secret. If I remember right you mentioned the yaml will get updated from the longhorn-test repro, where I committed the updated yaml?

yasker: I meant, the document should also show the code snippet for user, just like the minio-backuostore.yaml a few lines above. You can add cert section above, or show it here. Either is fine.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/backupstores-and-backuptargets.md
catherineluse: @yasker how should the grammar be fixed? should it be something like this?

> You will need to upload your own public- or private-CA-signed certificate and the associated key as PEM-encoded files. If you are using a certificate chain, all relevant certificates must be included by concatenating them.

yasker: @catherineluse Yes, this looks good. @joshimoo can you confirm?

joshimoo: @catherineluse @yasker I like the shorter text, but it's inaccurate since the user does not upload the certificate, instead it needs to be exposed as a AWS_CERT environment variable. Also it's important to understand that the public in this part refers to the certificate and has a special (loaded) meaning in the ssl/tls/https context, the same applies to the CA (certificate authority) statements above.

The reason I wrote such a long text is that the details are complicated, but if we don't need to be 100% accurate and specific, we certainly can go with some simplifications. :)

Here is some context information that might help: https://superuser.com/questions/620121/what-is-the-difference-between-a-certificate-and-a-key-with-respect-to-ssl

yasker: Is the following accurate?
> It's important to note that the certificate needs to be in PEM format and must either be a certificate authority, or a certificate chain concatenated with all relevant certificates, which includes all the certificate authorities.

joshimoo: I think we can go with your text, technically I think the or is referring to "the certificate" which would be inaccurate. since in the or case we are concatenating multiple certificates, etc together.

Pem itself is a container format that may include just the public certificate (such as with Apache installs, and CA certificate files /etc/ssl/certs), or may include an entire certificate chain including public key, private key, and root certificates.

yasker: After a lengthy discussion with @joshimoo , we've agreed on using the following sentence: 
> It's important to note that the certificate needs to be in PEM format, and must be it's own CA. Or one must include a certificate chain that contains the CA certificate. 
To include multiple certificates, one can just concatenate the different certificates (PEM files)

catherineluse:
@joshimoo can you please fix the merge conflict?

joshimoo:
@catherineluse It looks like you deleted the file i was working on? Is the new file for my content changes: set-backup-target.md?

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -122,6 +130,6 @@ The target URL should look like this:
 nfs://longhorn-test-nfs-svc.default:/opt/backupstore
 ```
 
-You can find an example NFS backupstore for testing purpose [here](https://github.com/longhorn/longhorn/blob/master/deploy/backupstores/nfs-backupstore.yaml). 
+You can find an example NFS backupstore for testing purpose [here](https://github.com/longhorn/longhorn/blob/master/deploy/backupstores/nfs-backupstore.yaml).
 
 **Result:** Longhorn can store backups in S3. To create a backup, see [this section.](../create-a-backup)
```
joshimoo, line 135: @catherineluse is wrong and should probably read in NFS?

joshimoo:
Moved my changes into set-backuptrarget... and rebased on top of staging.
@yasker @catherineluse  Can you check the final verbage?

yasker:
Link to https://github.com/longhorn/longhorn/issues/704

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -112,6 +113,13 @@ We provides two testing purpose backupstore based on NFS server and Minio S3 ser
...
 
+### Using a self signed SSL certificate for S3 communication
+If you want to use a self signed ssl certificate you can expose an `AWS_CERT` environment variable that contains your public certificate to the longhorn pods.
```
yasker, line 117: Is there another way to use custom CA other than set it in the secret? If not, we don't need to mention how we implemented it.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -112,6 +113,13 @@ We provides two testing purpose backupstore based on NFS server and Minio S3 ser
...
+We recommend you do so by adding the AWS_CERT key/value pair to the `minio-secret` or `aws-secret` and making it available to the `longhorn-system` namespace.
+For an example of using a self signed ssl certificate, have a look at our [minio-backupstore.yaml](https://github.com/longhorn/longhorn/blob/master/deploy/backupstores/minio-backupstore.yaml).
```
yasker, line 119: `self-signed SSL`

`or see the example in ` (insert anchor link to `Set up a Local Testing Backupstore`.)

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
yasker: Yeah, here should be `NFS` instead of `S3`.

catherineluse:
@joshimoo Yes, I deleted the file because the same content was being duplicated in two files. Thanks for remaking your changes to the other file.

I don't have any feedback, so if Sheng's feedback is addressed, LGTM.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
joshimoo: fixed also made the NFS change.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -112,6 +113,11 @@ We provides two testing purpose backupstore based on NFS server and Minio S3 ser
 
 **Result:** Longhorn can store backups in S3. To create a backup, see [this section.](../create-a-backup)
 
+### Using a self-signed SSL certificate for S3 communication
+If you want to use a self-signed ssl certificate you can expose an `AWS_CERT` environment variable that contains your public certificate to the longhorn pods, or see the example in [Set up a Local Testing Backupstore](#set-up-a-local-testing-backupstore).
```
yasker, line 117: `SSL` not `ssl`.

I don't feel we want to mention `environment variable` here. `AWS_CERT` is a part of the secret, not an environment variable. Using the secret's content as environment variables is how we implemented it internally, so the user doesn't need to know. Otherwise, they might be confused.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -112,6 +113,11 @@ We provides two testing purpose backupstore based on NFS server and Minio S3 ser
...
+It's important to note that the certificate needs to be in PEM format, and must be its own CA. Or one must include a certificate chain that contains the CA certificate.
+To include multiple certificates, one can just concatenate the different certificates (PEM files)
```
yasker, line 119: Missing ending period.

content/docs/0.8.1/snapshots-and-backups/backup-and-restore/set-backup-target.md
joshimoo: How about I remove all that and just go with see the example in...
As well as the note about the type, etc?

yasker: How about
> If you want to use a self-signed SSL certificate, you can specify `AWS_CERT` in the Kubernetes secret that you provided to Longhorn. See the example in [Set up a Local Testing Backupstore](#set-up-a-local-testing-backupstore).

yasker:
LGTM.

PR #106, opened 2020-05-15
Add root redirect for docs.longhorn.io

yasker:
Signed-off-by: Sheng Yang <sheng.yang@rancher.com>


yasker:
@lucperkins like this?

lucperkins:
That should work as far as I know

PR #116, opened 2020-05-25
Update maintenance.md

yasker:
Remove the content about Replica Rebuild Limit since we decided not to put it in v1.0.0.

Also addressed https://github.com/longhorn/longhorn/issues/1333


catherineluse:
LGTM

yasker:
LGTM

PR #122, opened 2020-05-28
Remove the instruction related to Longhorn v0.7.0

shuo-wu:
Since we won't support Longhorn v0.7.0 or lower versions after GA.

### DIFF
--- a/content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
+++ b/content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
@@ -5,8 +5,6 @@ weight: 1
 
 - [Upgrading from v0.8.1 to v1.0.0](#upgrading-from-v081-to-v100)
 - [Upgrading from v0.7.0+](#upgrading-from-v070)
-- [Upgrading from v0.6.2 or older version to v0.8.1](#upgrading-from-v062-or-older-version-to-v081)
-- [Upgrading from v0.6.2 to v0.7.0](#upgrading-from-v062-to-v070)
 
 ### Upgrading from v0.8.1 to v1.0.0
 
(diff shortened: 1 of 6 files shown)


content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
```diff
@@ -26,203 +24,3 @@ helm upgrade longhorn ./longhorn/chart
...
-
-- [Prerequisites](#prerequisites)
-- [Upgrade Longhorn Manager](#upgrade-longhorn-manager)
-  - [Using the Rancher Catalog App](#using-the-rancher-catalog-app)
-  - [Using kubectl](#using-kubectl)
-- [Troubleshooting](#troubleshooting)
-- [Clean up the v0.6.2 CRDs](#clean-up-the-v062-crds)
-- [Rollbacks](#rollbacks)
-
-### Prerequisites
```
yasker, line 99: This part of content is for general upgrade steps.

catherineluse:
So I should wait to merge it until after #122 ?

yasker:
@catherineluse hold on merging it. It removes the manager upgrade steps.

content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
shuo-wu: What... I think this part is for `v0.6.2 to v0.7.0` only. See line 87:
```
You will need to follow this guide to upgrade the Longhorn manager from v0.6.2 to v0.7.0.
```

Maybe we can retain some troubleshooting guides here. But we need to make sure they are for the general cases.

yasker: The steps here has more details than the steps in `### Upgrading Longhorn Manager from v0.7.0+`.

For example, the previous one doesn't mention upgrade using Rancher catalog app or have a list of the components that suppose to be running after the upgrade.

shuo-wu: I saw there are some updates for this doc. I decide not to modify this part for now.

yasker:
LGTM.

PR #129, opened 2020-05-29
Add more examples to docs 

shuo-wu:
longhorn/longhorn#1256

### DIFF
--- a/content/docs/0.8.1/advanced-resources/default-disk-and-node-config.md
+++ b/content/docs/0.8.1/advanced-resources/default-disk-and-node-config.md
@@ -89,3 +89,19 @@ If the label or annotation fails validation, the whole annotation is ignored.
 4. Wait for Longhorn to create the customized default disks automatically.
 
 > **Result:** The disks will be updated according to the annotation.
+
+### Launch Longhorn with multiple disks
+1. Add the label to all nodes before launching Longhorn. 
+
+    ```
+    node.longhorn.io/create-default-disk: 'config'
+    ```
+
+2. Then add the disk config annotation to all nodes:
+
+    ```
+    node.longhorn.io/default-disks-config: '[ { "path":"/var/lib/longhorn", "allowScheduling":true
+      }, { "name":"fast-ssd-disk", "path":"/mnt/extra", "allowScheduling":false, "storageReserved":10485760,
+      "tags":[ "ssd", "fast" ] }]'
+    ```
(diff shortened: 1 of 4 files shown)


shuo-wu:
```
9:04:11 PM: docs/0.8.1/deploy/upgrade/longhorn-manager/index.html
9:04:11 PM:   hash does not exist --- docs/0.8.1/deploy/upgrade/longhorn-manager/index.html --> ../../../references/settings/#replica-soft-anti-affinity
```

This check failure will be fixed by another PR

yasker:
You meant https://github.com/longhorn/website/pull/122 ?

shuo-wu:
> You meant #122 ? This one should be kept as an indirect link and has the name changed.

Yes. See https://github.com/longhorn/website/pull/122/files#diff-58ee1b2b386ee9bdd7029ce0818ae0a4L40

content/docs/0.8.1/references/settings.md
```diff
@@ -45,6 +45,8 @@ To configure Longhorn before installing it, see [this section](../../advanced-re
 #### Create Default Disk on Labeled Nodes
 > Example: `false`
 
+> Default: Un-checked
```
yasker, line 48: Let's remove `Example` and replace it with `Default` if there is a default value to the setting. If not, we can add an `Example` after the `Default` value. And for boolean type, use `true` or `false` is better since it's not only observed from the UI.

content/docs/0.8.1/advanced-resources/deploy/customizing-default-settings.md
```diff
@@ -82,10 +88,16 @@ From the project view in Rancher, go to **Apps > Launch > Longhorn** and edit th
       storageMinimalAvailablePercentage: 15
       upgradeChecker: false
       defaultReplicaCount: 2
-      guaranteedEngineCPU:
+      guaranteedEngineCPU: 0.25
```
yasker, line 91: Keep values of two examples the same.

content/docs/0.8.1/advanced-resources/default-disk-and-node-config.md
```diff
@@ -88,4 +88,18 @@ If the label or annotation fails validation, the whole annotation is ignored.
...
+1. Add the label to all nodes before launching Longhorn. 
+
+    ```
+    node.longhorn.io/create-default-disk: 'config'
+    ```
+
+2. Then add the disk config annotation to all nodes:
+
+    ```
+    node.longhorn.io/default-disks-config: <disks configuration with JSON string format>
```
yasker, line 103: Add a valid JSON string here. Don't need to manually format it to be pretty.

yasker:
LGTM

PR #131, opened 2020-05-29
doc: StorageClass and iscsi live upgrade

yasker:
https://github.com/longhorn/longhorn/issues/1429


catherineluse:
I couldn't modify this PR myself due to some credential issue but here's a couple of grammar suggestions

> We don't recommend modifying the default StorageClass named `longhorn`, since the change of parameters might cause issues during an upgrade later. If you want to change the parameters set in the StorageClass, you can create new StorageClass by referring to the [StorageClass examples](../references/examples/#storageclass).


> The `iSCSI` frontend does not support live upgrades.


> Note that only the "ext4" filesystem supports automatic remounts after a volume is detached unexpectedly. See [here](../../high-availability/recover-volume/) for details.

yasker:
@catherineluse Updated. Thanks for the review.

PR #136, opened 2020-05-30
Note a subpath may be used, with an example.

tedder:
..it works on 0.8, at least. I thought it might fail, so perhaps this will save some effort for others.

I haven't built the docs, I'm just assuming the format will be fine.


catherineluse:
Lgtm

yasker:
LGTM. Thanks @tedder .

PR #140, opened 2020-06-02
Fix the volume migration doc

shuo-wu:
longhorn/longhorn#1448

### DIFF
--- a/content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
+++ b/content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
@@ -48,9 +48,12 @@ Therefore, the PVCs and PVs should be migrated to use the new CSI plugin for the
     ```
     kubectl get pv --output=jsonpath="{.items[?(@.spec.csi.driver==\"io.rancher.longhorn\")].spec.csi.volumeHandle}"
     ```
-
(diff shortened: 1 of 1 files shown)


content/docs/0.8.1/deploy/upgrade/longhorn-manager.md
```diff
@@ -48,9 +48,12 @@ Therefore, the PVCs and PVs should be migrated to use the new CSI plugin for the
     ```
     kubectl get pv --output=jsonpath="{.items[?(@.spec.csi.driver==\"io.rancher.longhorn\")].spec.csi.volumeHandle}"
     ```
-
-2. Shut down the related workloads and detach the volumes. 
-3. Run this script for each volume:
+2. Remove finalizer `external-attacher/io-rancher-longhorn` for the related PV.
+    ```
+    kubectl edit pv <The corresponding PV of the volume found in step 1>
```
yasker, line 53: Can we use the patch command to remove the finalizer instead?

shuo-wu: Since there is another PV finalizer `kubernetes.io/pv-protection` handled by Kubernetes and `kubectl patch` would remove all finalizers only,  I prefer to use `kubectl edit` here.

PR #141, opened 2020-06-03
Add blog template

### LINKED ISSUE #111: Blog section
We want to add a blog section on the website for the community building effort.

We want it to be in time for the Longhorn v1.0.0 release.

lucperkins:
Fixes #111

### DIFF
--- a/assets/sass/style.sass
+++ b/assets/sass/style.sass
@@ -79,7 +79,7 @@ $code-background: $light
 .docs-content, .pagination
   +desktop-shrink
 
-.docs-content
+.has-extra-bottom-padding
   padding-bottom: 5rem
 
 .hero-logo
--- a/config.toml
+++ b/config.toml
@@ -84,10 +84,15 @@ sizes = [300, 400, 600, 700]
 type = "monospace"
 
 [[menu.main]]
+name = "Blog"
+url = "/blog"
+weight = 1
+
+[[menu.main]]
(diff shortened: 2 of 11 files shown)


yasker:
Thanks @lucperkins 

Notice the title `The Longhorn blog` should be `The Longhorn Blog`.

Also, can you remove the second blog post and rename the first one as `In Construction`? 

I need some advice on the scrolling down model (e.g. https://kubernetes.io/blog/ ) vs box model we have right now. Not sure which one is better. It seems most websites using the scrolling down model though.

lucperkins:
Those posts are just placeholders. They won’t show up in the live version of the site. When you write your first post, just delete them.

Also, could you say more about the scrolling model?

yasker:
@lucperkins I meant, it seems most website use scrolling down to show a list of blog posts ordered by the timestamp. The current blog layout is different. I think the scrolling down model is may be more straightforward for the blogs?

catherineluse:
I think that when you click Blog in the top navigation bar, it should take you to a list of the most recent blog posts. Right now, when I click blog, all I see is "The Rancher blog" and the second post isn't visible.

lucperkins:
@catherineluse I suspect you may not have applied the `--buildDrafts` flag.

yasker:
Merged, thanks @lucperkins !

PR #145, opened 2020-06-06
Change "ingress controller" to "Ingress"

oskapt:
"Ingress" and "ingress controller" are easily confused, since the Ingress is simply a set of instructions that the ingress controller uses to program an HTTP entrypoint. One does not create or modify an ingress controller after installation of the software into the cluster. One creates an Ingress, which is what the instructions on this page reflect.

These changes correct the language in this page of the docs.


yasker:
Thanks @oskapt to catch this. Can you add your `signed-off`? See https://github.com/longhorn/website#sign-off-on-all-commits

oskapt:
Superseded by #158.

PR #151, opened 2020-06-20
Add documentation for RWX workloads

joshimoo:
longhorn/longhorn#1183

Signed-off-by: Joshua Moody <joshua.moody@rancher.com>


content/docs/1.0.0/advanced-resources/rwx-workloads.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+{{< figure src="/img/diagrams/rwx/rwx-support.png" >}}
+
+Longhorn supports RWX workloads via the longhorn.io/nfs provisioner.
+To enable RWX support you need to deploy the following files:
+- [01-security.yaml](https://github.com/longhorn/longhorn/examples/rwx/01-security.yaml)
+- [02-longhorn-nfs-provisioner.yaml](https://github.com/longhorn/longhorn/examples/rwx/02-longhorn-nfs-provisioner.yaml)
+
+if you want to run multiple nfs-provisioners you can modify the provisioner argument and
```
shuo-wu, line 13: It would be better to remind users that scaling the deployment is not the correct way to launch multiple nfs-provisioners. Since there should be one unique StorageClass name for a NFS server (if I don't misunderstand `kubernetes_incubator/nfs-provisioner`).

yasker:
We need to include more information in this document. For example, what to expect when the node running nfs-provisioner went done, what if we have IP conflict for the service etc. Include all the things that the user might encounter while using this feature.

joshimoo:
updated the documentation, added an example with multiple nfs-provisioners.

yasker:
It's better if we can have one example to explain how to use the NFS provisioner here.

joshimoo:
rebased and added usage example (append only log)

content/docs/1.0.0/advanced-resources/rwx-workloads.md
```diff
@@ -0,0 +1,72 @@
+---
+title: RWX workloads
+weight: 4
+---
+
+# Longhorn RWX support
+
+{{< figure src="/img/diagrams/rwx/rwx-01.png" >}}
+
+Longhorn supports RWX workloads via the longhorn.io/nfs provisioner.
```
yasker, line 10: We've updated the provisioner's name.

joshimoo: fixed thanks

yasker:
I will merge this for now. @catherineluse Can you give it a second look when you have time? Thanks.

PR #154, opened 2020-06-29
docs/1.0.0/advanced-resources/deploy: remove the limitation warning about the length of the instance-manager image name

PhanLe1010:
Because we change from using a full image name to using a checksum for instance-manager labels, we remove the documentation about the limitation on instance-manager image name.

longhorn/longhorn#1323

### DIFF
--- a/content/docs/1.0.0/advanced-resources/deploy/airgap.md
+++ b/content/docs/1.0.0/advanced-resources/deploy/airgap.md
@@ -9,7 +9,6 @@ weight: 2
 
 #### Note:
   - CSI driver component's images, names and tags can be found [here.](../../../concepts/#13-csi-driver)
-  - We recommend using a short registry URL due to a Kubernetes limitation on the length of pod metadata labels. For more information, refer to [this section](./#longhorn-instance-manager-metadatalabels-must-be-no-more-than-63-characters)
 
 
 ## Using manifest file.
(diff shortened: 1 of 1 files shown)


yasker:
@PhanLe1010 This is for v1.0.1, so it should be in the staging branch.

@catherineluse Do you remember what we decided with the staging branch last time? We should have written it down but I forgot...

PhanLe1010:
@yasker I am not sure what to do with the staging branch (create a new folder for v1.0.1 docs, or something else ...) I will wait for @catherineluse 's response before I proceed.

catherineluse:
@yasker After the conceptual docs were merged into master, I expected that staging would be identical to master. But it appears that there is still a large diff. https://github.com/longhorn/website/pull/155 do you see why?

@PhanLe1010 When staging is up to date with master, make changes to the 1.0.0 folder of the docs in the staging branch. Then when v1.0.1 is released, the following will happen:

- The 1.0.0 folder will be renamed to v.1.0.1 in staging
- The 1.0.0 will be copied from master
- Then staging will merge into master

content/docs/1.0.0/advanced-resources/deploy/airgap.md
```diff
@@ -286,48 +285,6 @@ weight: 2
...
-    longhorn.io/component: instance-manager
-    longhorn.io/instance-manager-image: <PRIVATE_REGISTRY_URL>-longhornio-longhorn-instance-manager-v1_20200301
-    longhorn.io/instance-manager-type: engine
-    longhorn.io/node: <NODE_NAME>
-  name: instance-manager-e-XXXXXXXX
-```
-
-it's known Kubernetes limitaion that label value should be no more than 63 characters [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
-
 ##### Recommendation:
```
shuo-wu, line 288: This recommendation is part of the section `Longhorn instance manager: metadata.labels must be no more than 63 characters`. Make it an independent section.

PhanLe1010:
Thanks @catherineluse for the instruction. I see that @yasker already changed the base branch to `staging`. So, I think that I only need to fix the `change request of @shuo-wu

content/docs/1.0.0/advanced-resources/deploy/airgap.md
PhanLe1010: Thank you. Fixed

PR #164, opened 2020-07-10
Update docs with Helm repo

ttpcodes:
This PR updates the docs on the website to reference the new `Helm` repository at https://charts.longhorn.io. The `Helm` installation instructions have been modified to reference that repository instead of having the user clone the `Longhorn` repository to access the chart. Additionally, the instructions for modifying the `Default Settings` in `Longhorn` have been updated to omit the `git clone` step and to indicate how to obtain and modify a copy of `values.yaml` for `Helm` if using the `--values` flag.


yasker:
LGTM. @catherineluse ?

PR #165, opened 2020-07-14
Add instruction to enable virtual-hosted-style access for S3 backupstore

shuo-wu:
longhorn/longhorn#1552


shuo-wu:
Do we need to write concrete instructions for Aliyun OSS like AWS S3? But I don't know how to config the IAM user (See [step 2](https://longhorn.io/docs/1.0.0/snapshots-and-backups/backup-and-restore/set-backup-target/#set-up-aws-s3-backupstore)) in Aliyun...

yasker:
> Do we need to write concrete instructions for Aliyun OSS like AWS S3? But I don't know how to config the IAM user (See [step 2](https://longhorn.io/docs/1.0.0/snapshots-and-backups/backup-and-restore/set-backup-target/#set-up-aws-s3-backupstore)) in Aliyun...

That's fine. I think the users of Aliyun should be able to figure it out.

shuo-wu:
Can we merge this PR?

PR #168, opened 2020-07-16
Add a doc explain the node space usage info

shuo-wu:
longhorn/longhorn#1519


yasker:
The content looks good. The first picture is weird though. You can check it by looking into the preview page.

Screen Shot 2020-07-17 at 9 35 41 AM

shuo-wu:
> The content looks good. The first picture is weird though. You can check it by looking into the preview page.
> 
> Screen Shot 2020-07-17 at 9 35 41 AM

I am not sure which part is weird. But I still updated the picture to present more info.

yasker:
@shuo-wu where is the middle circle?...

shuo-wu:
The picture has been updated.

PR #169, opened 2020-07-20
Add a doc explains the volume size related concepts

shuo-wu:
longhorn/longhorn#1519


shuo-wu:
Need to wait for `node-space-usage.md` merge then re-run the check for this PR.

PR #172, opened 2020-07-21
Add a doc explains the volume size related concepts

shuo-wu:
longhorn/longhorn#1519


shuo-wu:
For the screenshots in the example, I am not sure if I crop too much of the pictures.

PR #173, opened 2020-07-27
Add note about root requirement to deploy docs

ttpcodes:
This PR adds a note to the list of requirements for deploying Longhorn indicating that Longhorn needs root access in order for Longhorn to be properly deployed and operated as requested in longhorn/longhorn#1625. Documentation about Pod Security Policy requirements has been omitted due to concerns about Pod Security Policy being deprecated in future versions of Kubernetes.

### DIFF
--- a/content/docs/1.0.1/deploy/install/_index.md
+++ b/content/docs/1.0.1/deploy/install/_index.md
@@ -66,6 +66,12 @@ daemonset.apps "longhorn-environment-check" deleted
 clean up complete
 ```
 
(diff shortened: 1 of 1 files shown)


yasker:
We haven't made a decision on PSP yet. Put this on hold until we make the decision to ship with PSP by default or not.

yasker:
@ttpcodes Since we've agreed on to ship with PSP, can you update the PR with the PSP part?

PR #174, opened 2020-07-29
Update create-volumes.md

just1689:
Bug fix - the file referenced does not exist and returns a 404. I have check and replaced it with the file that does exist.


yasker:
Thanks for the PR @just1689 .

Can you sign-off the commit? See https://github.com/longhorn/website#sign-off-on-all-commits for details.

PR #182, opened 2020-08-17
Update version to v1.0.2

vflong:
Signed-off-by: weifeilong <weifeilong2013@gmail.com>


yasker:
Thanks for catching these!

PR #191, opened 2020-09-07
Update quick-start-with-helm.md  

coolamiy:
updated  helm comamnds for helm2 and helm3 for helm2 and helm3 versions on the page.


yasker:
Hi @coolamiy 

Thanks for the PR. Can you update it to be consistent with v0.8.1? The wordings are slightly different. See https://longhorn.io/docs/0.8.1/deploy/install/install-with-helm/

yasker (approved):

Issue #197, opened 2020-09-25
Add Knowledge Base section to the website


yasker:
https://github.com/longhorn/longhorn/issues/1783

So we can start answering the most common questions, as well as provide the doc as the source instead of asking users to check the GitHub issue thread.

hackabletype:
Hi @yasker, I can take a look at this.

Should this work more like the current Blog page (a collection of pages each containing a KB item), or the Project page (a single page with a list of all the KB items), or should it more closely mimic the actual docs architecture? Or did you have something else entirely in mind?

yasker:
Thanks @hackabletype . I think it would be similar to the current Blog pages since KB articles are less organized than the doc. As long as it's easy for the user to search through the whole repo, it should be fine.

yasker:
Done by https://github.com/longhorn/website/pull/198

PR #198, opened 2020-09-30
Adding Knowledge Base page, including content example

hackabletype:
As per issue https://github.com/longhorn/website/issues/197, adding a Knowledge Base page based on the existing architecture of the Blog pages.

Open questions:
- [ ] How do these new pages/sections get added to the Algolia search
- [x] Should this be merged into longhorn/website master branch, or the v1.1.0 branch
- [x] Is there a better content example


zacharysarah:
Deploy preview: https://deploy-preview-198--longhornio.netlify.app/kb/

celestehorgan:
Comments and questions:

1) If you intend for a page to use the same base layouts as defined in `/layouts/_default`, you don't need to copy them into their own subdirectory, i.e. `/layouts/kb`. Hugo uses a fallback model for layouts, so if it doesn't find any layout files in `/layouts/kb`, it automatically falls back to `/layouts/_default`. 

Follow up question #1: Is the intent to customize the layout of this page?

2) Page structure: is each question going to have its own page, or be a list on _index.md?

3) Re: content examples – this is a knowledge base, I think you can assume each content node is going to be a question-answer pair.

4) Algolia search – this is done via whomever has the login for Algolia for this site, which is not me.

5) Branches: this is up to the maintainers of this project, so please @-mention one for review.


----


Ping me when you've gotten a little further with this :)

hackabletype:
@celestehorgan makes a good suggestion, especially since I don't think I can add reviewers using the UI.

@yasker, could you take a look at this PR (and the couple of questions I've got)?

hackabletype:
@celestehorgan:
> Follow up question #1: Is the intent to customize the layout of this page?

No, my understanding is that I should be like the existing blog pages.

hackabletype:
I reverted the commit where I removed the layouts/kb files for two reasons:
1. The change broke the deploy-preview
2. I think we actually do need them. Without them the Knowledge Base pages don't get formatted as the Blog pages do.

yasker:
@hackabletype Thanks for the PR. It looks good!

1. Regarding the Algolia search, it was @lucperkins who helped to add it IIRC. I've just tried, it seems the blog wasn't covered by the Algolia search as well (but Google can find it). @caniszczyk do we have an account with Algolia? 
2. Merge into Longhorn master is fine, since v1.1.0 will merge from master as well.
3. Regarding the format, we can use something like this

```
# Title

## Details

## Resolution

## Related information
```

We should able to put the first post there as a draft, then I can add the real content to it.

hackabletype:
https://deploy-preview-198--longhornio.netlify.app/

@yasker I've made the content example change you've suggested. I'm not able to assign or merge, so if you're happy with this PR I think you'll need to do the merge.

I think if the Algolia also affects the Blog pages, then maybe we should take the Algolia section of [your comment](https://github.com/longhorn/website/pull/198#issuecomment-701710159) and make it a separate issue.

yasker:
@hackabletype Sounds good. I will merge the PR then. Thanks!

PR #202, opened 2020-10-02
Update concepts.md

yasker:
Fixes https://github.com/longhorn/longhorn/issues/1682 and https://github.com/longhorn/longhorn/issues/1772

Signed-off-by: Sheng Yang <sheng.yang@rancher.com>


yasker:
Close in favor of #203

PR #222, opened 2020-11-06
Mention that any Kubernetes-compatible container runtime is sufficient

psy-q:
Longhorn does not specifically require Docker. I went by these requirements and later found out that any Kubernetes container runtime works through [a discussion on Slack](https://rancher-users.slack.com/archives/CC2UQM49Y/p1604669734311000).

### DIFF
--- a/content/docs/0.8.0/getting-started/quick-start-with-helm.md
+++ b/content/docs/0.8.0/getting-started/quick-start-with-helm.md
@@ -8,7 +8,7 @@ weight: 5
 ## Quick Start
 
 1. Helm 3.0+
-2. Docker v1.13+
+2. A container runtime compatible with Kubernetes (Docker v1.13+, containerd v1.3.7+, etc.)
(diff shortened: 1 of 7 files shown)


yasker:
@psy-q Thanks for the PR.

Can you squash the commits into one? Also, we need your sign off on the commit. See https://github.com/longhorn/longhorn/blob/master/CONTRIBUTING.md#dco-sign-off

And can you update the doc for other versions as well? It should be applied to every Longhorn version so far. At least we want to have it in v1.0.2 doc (which is our latest doc).

psy-q:
Oh, sorry, I didn't notice the other versions before, I just hunted around for that requirements.md. I now went ahead and changed it for all versions and squashed the commits, I hope this is OK.

yasker (approved):

yasker:
Merged, thanks @psy-q !

PR #228, opened 2020-11-22
requirements.md: further explain open-iscsi req

flokli:
See https://github.com/longhorn/longhorn/issues/1984

### DIFF
--- a/content/docs/0.8.0/install/requirements.md
+++ b/content/docs/0.8.0/install/requirements.md
@@ -5,11 +5,11 @@ weight: 1
 
 -  A container runtime compatible with Kubernetes (Docker v1.13+, containerd v1.3.7+, etc.)
 -  Kubernetes v1.14+.
--  `open-iscsi` has been installed on all the nodes of the Kubernetes cluster, and `iscsid` daemon is running on all the nodes.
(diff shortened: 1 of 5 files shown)


flokli:
cc @yasker

content/docs/0.8.0/install/requirements.md
```diff
@@ -5,7 +5,7 @@ weight: 1
 
 -  A container runtime compatible with Kubernetes (Docker v1.13+, containerd v1.3.7+, etc.)
 -  Kubernetes v1.14+.
--  `open-iscsi` has been installed on all the nodes of the Kubernetes cluster, and `iscsid` daemon is running on all the nodes.
+-  `open-iscsi` has been installed on all the nodes of the Kubernetes cluster, and `iscsid` daemon is running on all the nodes. This is necessary, as all longhorn volumes are created as iscsi targets under the hood.
```
yasker, line 8: How about
```
Since Longhorn relies on `iscsiadm` on the host to provide persistent volumes to Kubernetes.
```

Also, we can update the latest doc (v1.0.2, instead of v0.8.0) and it will be picked up by the upcoming v1.1.0 release as well.

flokli: off, not sure how I got to 0.8.0 only. I updated all docs according to your suggestion, PTAL.

yasker (approved):

yasker:
Merged, thanks @flokli !

PR #229, opened 2020-11-23
Update the documents for private registry secret name

PhanLe1010:
It is moved from default setting to private registry section

longhorn/longhorn#1670


content/docs/1.1.0/advanced-resources/deploy/airgap.md
```diff
@@ -293,16 +293,14 @@ If you want to use custom images' names, you can use the following steps:
 
 If you keep the images' names as recommended [here](./#recommendation), you only need to do the following steps:
 
-- In the `Longhorn Default Settings` section, click **Customize Default Setting** and specify the Private registry secret.
 - In the `Private Registry Settings` section specify:
    - Private registry URL
    - Private registry user
    - Private registry password
+   - Private registry secret name
```
yasker, line 300: Are we going to generate the secret automatically according to the questionnaire? If so, we should call it out, otherwise, the user will be confused about what to fill in here.

yasker (changes requested):

content/docs/1.1.0/advanced-resources/deploy/airgap.md
PhanLe1010: Thank you. Fixed it

yasker (approved):

PR #247, opened 2020-12-30
Fix pagination in Docs

bk201:
Navigate between section pages instead of all regular pages.

Signed-off-by: Kiefer Chang <kiefer.chang@suse.com>

The PR fixes the problem that the previous and next buttons link to the wrong pages, e.g.

Screen Shot 2020-12-30 at 12 05 24 PM


yasker:
Merged, thanks @bk201 !

PR #251, opened 2021-01-07
Use release tag to install/upgrade instead of master branch

jenting:
Use [shortcode](https://gohugo.io/content-management/shortcodes/) to render the current doc version in the markdown template.

To get the current version, use template `{{- index (split .Page.File.Dir "/") 1 -}}` in `layouts/shortcodes/current-version.html`
Then for each markdown file, use template `v{{< current-version >}}` to render the version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@ This repo houses the assets used to build the website for Longhorn, available at
 To run the website locally, you need to have the [Hugo](https://gohugo.io) static site generator installed (installation instructions [here](https://gohugo.io/getting-started/installing/)). Once Hugo is installed:
 
 ```bash
+npm install bulma
 hugo server --buildDrafts --buildFuture
 ```
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This repo houses the assets used to build the website for Longhorn, available at
 To run the website locally, you need to have the [Hugo](https://gohugo.io) static site generator installed (installation instructions [here](https://gohugo.io/getting-started/installing/)). Once Hugo is installed:
 
 ```bash
-npm install bulma
+yarn install
 hugo server --buildDrafts --buildFuture
 ```
 
--- a/content/docs/0.8.0/install/install-with-kubectl.md
+++ b/content/docs/0.8.0/install/install-with-kubectl.md
@@ -9,7 +9,7 @@ weight: 8
 You can install Longhorn on any Kubernetes cluster using this command:
 
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/longhorn/longhorn/master/deploy/longhorn.yaml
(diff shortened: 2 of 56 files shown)


README.md
```diff
@@ -9,6 +9,7 @@ This repo houses the assets used to build the website for Longhorn, available at
 To run the website locally, you need to have the [Hugo](https://gohugo.io) static site generator installed (installation instructions [here](https://gohugo.io/getting-started/installing/)). Once Hugo is installed:
 
 ```bash
+npm install bulma
```
yasker, line 12: I don't recall we need to install `bulma` before. Is it something new?

yasker (changes requested):
LGTM in general. And yes, I'd love to have a solution to automatically replace the release tag on doc directories. Maybe a script can help.

README.md
jenting: I can't run `hugo server --buildDrafts --buildFuture` or `make serve` success on a pure docker container (ubuntu:focal _or_ opensuse/leap). The error message is
```
Error: Error building site: TOCSS: failed to transform "sass/style.sass" (text/x-sass): SCSS processing failed: file "stdin", line 19, col 1: File to import not found or unreadable: bulma/sass/utilities/initial-variables.
make: *** [Makefile:2: serve] Error 255
```

I'm not sure is it something new but after `npm install bulma`, local build success.

yasker: I see. Sounds like previously the dependency was installed on the host.

yasker:
@jenting Can you file an issue to track the automatic replacement of the version tag for the doc in `longhorn/longhorn`? Just to make sure we will remember to fix it. After that, we can merge this PR.

jenting:
I found a way to automatically replace the version.

I've separate different doc version change to different git commits to helping reviewer review.
Please take a look :bow:

README.md
bk201: How about just

```
yarn
```

Since there is `yarn.lock` file.

README.md
```diff
@@ -9,7 +9,7 @@ This repo houses the assets used to build the website for Longhorn, available at
 To run the website locally, you need to have the [Hugo](https://gohugo.io) static site generator installed (installation instructions [here](https://gohugo.io/getting-started/installing/)). Once Hugo is installed:
 
 ```bash
-npm install bulma
+yarn
```
yasker, line 12: Just `yarn` is good enough? I don't know much about `npm` vs `yarn` though.

yasker (changes requested):
This looks great. Thanks for finding the way to automate it!

LGTM in general.

README.md
jenting: Because the package.json and yarn.lock exist, _yarn_ will look up the package.json and install the dependency. 

Command "yarn" or "yarn install" both work, ref to https://yarnpkg.com/lang/en/docs/usage/

yasker (approved):

PR #268, opened 2021-03-09
update longhorn ingress manifest from v1beta1 to v1

cclhsu:
update longhorn ingress manifest from v1beta1 to v1 to resolve deprecated warning message during apply/delete.

Signed-off-by: Clark Hsu <clark.hsu@suse.com>


innobead:
> update longhorn ingress manifest from v1beta1 to v1 to resolve deprecated warning message during apply/delete.
> 
> 
> 
> Signed-off-by: Clark Hsu <clark.hsu@suse.com>

Small fix is ok w/o issue in general, but here I would suggest to have an issue mapping to any PR as a basic convention.

PR #280, opened 2021-04-02
Update document for taint tolerations settings

PhanLe1010:
User need to set tolerations for system components and user-deployed
components separately

longhorn/longhorn#2120


shuo-wu (approved):
LGTM

innobead (approved):

PR #290, opened 2021-04-21
Add Kb - Recurring job does not create new jobs after detach and attach

c3y1huang:
https://github.com/longhorn/longhorn/issues/2513

### DIFF
--- a/content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
+++ b/content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
@@ -1,0 +1,52 @@
+---
+title: "Troubleshooting: Recurring job does not create new jobs after detaching and attaching volume"
+author: Chin-Ya Huang
+draft: false
+date: 2021-04-21
+categories:
+  - "recurring job"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
+Recurring job does not create new jobs when the volume is attached after being detached for a long time.
+
+According to Kubernetes [CronJob limitations](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations):
+> For every CronJob, the CronJob Controller checks how many schedules it missed in the duration from its last scheduled time until now. If there are more than 100 missed schedules, then it does not start the job and logs the error
+>
+>```
+>Cannot determine if job needs to be started. Too many missed start time (> 100). Set or decrease .spec.startingDeadlineSeconds or check clock skew.
+>```
+
+That means the duration of the attach/detach operation that the recurring job can tolerate is depending on the scheduled interval.
+
+For example, if the recurring backup job is set to run every minute, then the toleration would be 100 minutes.
+
+## Solution
+
+Directly delete the stuck cronjob to allow Longhorn to recreate it.
+```
+ip-172-30-0-211:/home/ec2-user # kubectl -n longhorn-system get cronjobs
+NAME                                                  SCHEDULE    SUSPEND   ACTIVE   LAST SCHEDULE   AGE
+pvc-394e6006-9c34-47da-bf27-2286ae669be1-c-ptl8l1-c   * * * * *   False     1        47s             2m23s
+
+ip-172-30-0-211:/home/ec2-user # kubectl -n longhorn-system delete cronjobs/pvc-394e6006-9c34-47da-bf27-2286ae669be1-c-ptl8l1-c
+cronjob.batch "pvc-394e6006-9c34-47da-bf27-2286ae669be1-c-ptl8l1-c" deleted
+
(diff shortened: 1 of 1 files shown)


content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
```diff
@@ -0,0 +1,52 @@
...
+author: Chin-Ya Huang
+draft: false
+date: 2021-04-21
+categories:
+  - "recurring job"
+---
+
+## Applicable versions
+
+All Longhorn versions.
```
PhanLe1010, line 12: Should we indicate `Version v1.1.1 and earlier` ? So, we fix this in v1.1.2 we don't have to modify the KB

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Symptoms
+
+Recurring job does not create new jobs when the volume is attached after being detached for sometimes.
```
PhanLe1010, line 16:
```suggestion
Recurring job does not create new jobs when the volume is attached after being detached for a long time.
```

PhanLe1010 (commented):
LGTM.
Still need @catherineluse to review

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
c3y1huang: I have no problem having the version pre-set before the fix, but thought it might make more sense to be in sync with the current status. @innobead , can we get some opinion on this?

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
c3y1huang: Thanks, addressed.

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
innobead: I think we can revisit back later when we really fix the issue. @PhanLe1010, update this action item in the related ticket? Thanks.

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
```diff
@@ -0,0 +1,52 @@
...
+Recurring job does not create new jobs when the volume is attached after being detached for a long time.
+
+According to Kubernetes [CronJob limitations](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations):
+> For every CronJob, the CronJob Controller checks how many schedules it missed in the duration from its last scheduled time until now. If there are more than 100 missed schedules, then it does not start the job and logs the error
+>
+>```
+>Cannot determine if job needs to be started. Too many missed start time (> 100). Set or decrease .spec.startingDeadlineSeconds or check clock skew.
+>```
+
+The means duration of the attach/detach that the recurring job can tolerate is depending on the scheduled interval.
```
innobead, line 25:
```suggestion
That means the duration of the attach/detach operation that the recurring job can tolerate is depending on the scheduled interval.
```

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
```diff
@@ -0,0 +1,52 @@
...
+
+For example, If the recurring backup job is set to run every minute, then the toleration would be 100 minutes.
```
innobead, line 27:
```suggestion
For example, if the recurring backup job is set to run every minute, then the toleration would be 100 minutes.
```

innobead (changes requested):
LGTM, minor changes requested

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
```diff
@@ -0,0 +1,52 @@
+---
+title: "Troubleshooting: Recurring job does not create new jobs after detach and attach"
```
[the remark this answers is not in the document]
c3y1huang: Fixed.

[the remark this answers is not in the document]
c3y1huang: Fixed.

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
c3y1huang: Fixed.

content/kb/troubleshooting-recurring-job-stuck-after-detach-attach.md
c3y1huang: Fixed.

innobead (approved):
LGTM

PR #296, opened 2021-04-28
Add Kb - troubleshoot none-standard Kubelet directory

c3y1huang:
Feel like we should add this to kb.

https://github.com/longhorn/longhorn/issues/2537

### DIFF
--- a/content/kb/troubleshooting-none-standard-kubelet-dir.md
+++ b/content/kb/troubleshooting-none-standard-kubelet-dir.md
@@ -1,0 +1,62 @@
+---
+title: "Troubleshooting: None-standard Kubelet directory"
+author: Chin-Ya Huang
+draft: false
+date: 2021-06-18
+categories:
+  - "csi"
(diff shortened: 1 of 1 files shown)


content/kb/troubleshooting-volume-mount-with-none-standard-kubelet-dir.md
```diff
@@ -0,0 +1,39 @@
...
+  - "csi"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
+When the Kubernetes cluster is using a non-standard Kubelet directory, PVC gets created, attached, but never mounted.
```
shuo-wu, line 16: Can the volume be created or attached?

c3y1huang:
Convert to draft until https://github.com/longhorn/longhorn/issues/2537 is figured out.

innobead (approved):

PR #298, opened 2021-04-29
Add KB - default setting not persist after upgrade

c3y1huang:
https://github.com/longhorn/longhorn/issues/2542

### DIFF
--- a/content/kb/troubleshooting-default-settings-do-not-persist.md
+++ b/content/kb/troubleshooting-default-settings-do-not-persist.md
@@ -1,0 +1,31 @@
+---
+title: "Troubleshooting: Longhorn default settings do not persist"
+author: Chin-Ya Huang
+draft: false
+date: 2021-04-29
+categories:
+  - "setting"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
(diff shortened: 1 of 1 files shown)


content/kb/troubleshooting-default-setting-not-persist-after-upgrade.md
```diff
@@ -0,0 +1,29 @@
...
+  - "setting"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
+I have modified Longhorn default settings, but they do not persist after upgrade.
```
shuo-wu, line 16: Add more details here. e.g., When upgrading Longhorn system via helm or Rancher App, the modified Longhorn default settings doesn't work... When modifying the default settings via `kubectl -n longhorn-system edit configmap longhorn-default-setting`, the modification won't be applied to Longhorn system.

shuo-wu (changes requested):

content/kb/troubleshooting-default-setting-not-persist-after-upgrade.md
c3y1huang: Thank you. Fixed.

shuo-wu (approved):

innobead (approved):

PR #299, opened 2021-05-05
Warn backup restore on XFS filesystem

c3y1huang:
https://github.com/longhorn/longhorn/issues/2503


yasker:
I am wondering, should we release a patch release to fix this? Along with other high-priority bugs.

innobead (approved):
LGTM

PR #301, opened 2021-05-14
Make homepage elegant

jenting:
WAS:
image

IS:
image


innobead (changes requested):
Don't change project-info.html, because I think the style is still good. Just need to modify footer.html.

innobead (approved):

PR #304, opened 2021-05-14
Fix search partial work

jenting:
The anchor of the class name configured in Algolia is `doc-content`, not `container`.
Ref to https://github.com/algolia/docsearch-configs/blob/680a4c9/configs/longhorn.json#L16-L21

Run the test locally with https://docsearch.algolia.com/docs/run-your-own.
The search result of `nb_hit` increases from 32 to 15649.

The Algolia crawler run every 24 hours https://docsearch.algolia.com/docs/what-is-docsearch#quick-description
So the production Algolia index would be update after this PR merged within 24 hours.

#### Issue

https://github.com/longhorn/longhorn/issues/2155

### DIFF
--- a/layouts/partials/blog/content.html
+++ b/layouts/partials/blog/content.html
@@ -1,6 +1,6 @@
 {{ $posts := where site.RegularPages "Section" "blog" }}
 <section class="section">
-  <div class="container">
+  <div class="docs-content">
     <div class="columns">
       <div class="column is-two-thirds">
         <div class="content is-medium has-extra-bottom-padding">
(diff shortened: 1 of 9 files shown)


jenting:
The next is to contact Algolia support engineer to update the Longhorn config.json to allow filter by version
```
  "custom_settings": {
    "attributesForFaceting": [
      "version"
    ]
  },
```

and in Longhorn website, add `algoliaOptions` in Algolia request API.
```html
<script type="text/javascript"> docsearch({
    ...
    indexName: 'longhorn',
    inputSelector: '#search-input',
    algoliaOptions: {'facetFilters': ["version:{{ <current-version> }}"]},
    ...
  });
</script>
```

innobead (approved):

Issue #305, opened 2021-05-18
Algolia search doesn't search current document version


jenting:
Currently, if you input a search keyword by Algolia, the result includes all the document versions.
It's better to filter the result by the current document version.

innobead:
@jenting it’s better to have some description instead of empty.

PR #306, opened 2021-05-18
Add version facet filter

jenting:
#### Issue

#305

This depends on https://github.com/algolia/docsearch-configs/pull/4090

### DIFF
--- a/layouts/partials/javascript.html
+++ b/layouts/partials/javascript.html
@@ -28,12 +28,14 @@
 </script>
 
 {{/* Algolia search */}}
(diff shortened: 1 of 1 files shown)


innobead (changes requested):
If no content, u can use <script ... /> directly instead of a separate </script>

jenting:
> If no content, u can use <script ... /> directly instead of a separate </script>

Hmmm... I did not want to change that line.
BTW, I tried using `<script ... />` directly, however, the Javascript doesn't work anymore.

innobead (approved):

PR #315, opened 2021-05-28
Support replica auto-balance

c3y1huang:
https://github.com/longhorn/longhorn/issues/587

### DIFF
--- a/content/docs/1.2.0/high-availability/auto-balance-replicas.md
+++ b/content/docs/1.2.0/high-availability/auto-balance-replicas.md
@@ -1,0 +1,106 @@
+---
+  title: Auto Balance Replicas
+  weight: 1
+---
+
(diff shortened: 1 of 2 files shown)


shuo-wu (commented):
In general LGTM.
Do we need to provide more special cases then explain the criteria in detail? e.g., the case in test `test_replica_auto_balance_zone_best_effort_with_uneven_node_in_zones`. This case means Longhorn will try to guarantee no duplicate replica on each node first rather than forcibly making replicas evenly distributed in 2 zones.

c3y1huang:
> Do we need to provide more special cases then explain the criteria in detail? e.g., the case in test `test_replica_auto_balance_zone_best_effort_with_uneven_node_in_zones`. This case means Longhorn will try to guarantee no duplicate replica on each node first rather than forcibly making replicas evenly distributed in 2 zones.

I think we should, thank you and addressed.

shuo-wu (approved):

innobead (approved):

PR #329, opened 2021-06-22
Add a note for AWS IAM Role ARN

jenting:
The kube2iam/kiam Pods must be the same node as longhorn-manager as well
as replica instance manager Pods.

#### Linked Issue

https://github.com/longhorn/longhorn/issues/2718


jenting:
@innobead PTAL

innobead (approved):

PR #350, opened 2021-08-12
Add doc for async pull backup target

jenting:
#### Issues
https://github.com/longhorn/longhorn/issues/1761


content/docs/1.2.0/snapshots-and-backups/backup-and-restore/_index.md
```diff
@@ -1,4 +1,10 @@
 ---
 title: Backup and Restore
 weight: 2
----
\ No newline at end of file
+---
+
```
innobead, line 5: I think the statement here can be as a note by using `>` markdown syntax because it is describing the change between before and after 1.2.0.

Below this new note, you can use a TOC of the subpages to enrich this page like below monitoring page.

image

innobead (changes requested):
LGTM, but suggest some changes.

innobead (approved):
LGTM

PR #353, opened 2021-08-16
Update Recurring Snapshots and Backups

c3y1huang:
https://github.com/longhorn/longhorn/issues/467


innobead:
@c3y1huang LGTM, but we need a upgrade note in the upgrade note page to inform users there are some recurring jobs migration happening during the upgrade.

c3y1huang:
> @c3y1huang LGTM, but we need a upgrade note in the upgrade note page to inform users there are some recurring jobs migration happening during the upgrade.

Note added.

innobead (approved):
LGTM

PR #365, opened 2021-08-30
Update ingress example for uploading backing image

innobead:

https://github.com/longhorn/longhorn/issues/2937


kaxing:
Reviewed, this setting works on my test cluster.

A note, in nginx it seems the `client_max_body_size` can be set to 0 to bypass the checking, but this need to be test again.

Ref.:
*https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size
**http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

kaxing (approved):
Reviewed, this setting works on my test cluster.

A note for possible improvment, in nginx it seems the `client_max_body_size` can be set to 0 to bypass the checking, but this need to be test again.

Ref.:
*https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-max-body-size
**http://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size

PR #367, opened 2021-08-31
Make sure user understand the default installation process is not for production usage

PhanLe1010:
longhorn/longhorn#2859


innobead (approved):
LGTM

PR #372, opened 2021-09-01
Add slack button

jenting:
#### Proposal Change

- Add slack at the right top
- Remove the unnecessary GitHub image file

Screenshot from 2021-09-01 10-06-56

#### Issue

https://github.com/longhorn/longhorn/issues/2875


config.toml
```diff
@@ -58,6 +58,7 @@ If your main cluster fails, you can bring up the app in the DR cluster quickly w
 
 [params.social]
 github = "https://github.com/longhorn/longhorn"
+slack = "https://slack.cncf.io"
```
yasker, line 61: This won't redirect directly to Longhorn channel right? I am wondering if there is a better way.

jenting: This would work https://cloud-native.slack.com/messages/longhorn

yasker (approved):

PR #375, opened 2021-09-14
Use the `latest` url in canonical tag

jenting:
To make the search engine do not create duplicate indexing across multiple doc versions, use the `latest` URL in the canonical tag. For example:
- was
```html
<link rel="canonical" href="https://longhorn.io/docs/1.2.0/deploy/install/" />
```
- is
```html
<link rel="canonical" href="https://longhorn.io/docs/latest/deploy/install/" />
```
So, the search engine will only index that page instead of all duplicated version pages.

https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls#rel-canonical-link-method

https://github.com/longhorn/longhorn/issues/2758


innobead (approved):
LGTM

PR #378, opened 2021-09-21
Update the StorageClass example and the auto remount description

shuo-wu:
https://cloud-native.slack.com/archives/CNVPEL9U3/p1632123916152100?thread_ts=1631890074.139400&cid=CNVPEL9U3


innobead (approved):
LGTM

PR #387, opened 2021-10-04
Remove redundant word

derekbit:


innobead:
@derekbit Please help update all versions of the doc instead of only 1.0.2. Thanks!

PR #393, opened 2021-10-06
Fix Markdown for RecurringJob definition

Pumba98:
Hi, I know, huge change 😄 But docs for 1.2.1 are unreadable here https://longhorn.io/docs/1.2.1/snapshots-and-backups/scheduling-backups-and-snapshots/#set-up-recurring-jobs-using-a-longhorn-recurringjob


innobead (approved):
LGTM

innobead:
@Pumba98 thanks for the fix!

PR #409, opened 2021-11-04
Remove broken link of Rancher doc

jenting:
Rancher doc separates to use different URL links, such as v2.0-v2.4, v2.5, v2.6.
To make Longhorn documentation compatible across all Rancher docs, I think we should remove the hyperlink directly.
Then users have to find the related doc by themselves.

https://github.com/longhorn/longhorn/issues/3215


innobead (approved):
LGTM

PR #410, opened 2021-11-08
Add KB doc for volume cleanup issue after node is down and recovered

derekbit:
https://github.com/longhorn/longhorn/issues/3080

### DIFF
--- a/content/kb/troubleshooting-volume-cannot-be-cleaned-up-after-the-node-of-the-workload-pod-is-down-and-recovered.md
+++ b/content/kb/troubleshooting-volume-cannot-be-cleaned-up-after-the-node-of-the-workload-pod-is-down-and-recovered.md
@@ -1,0 +1,25 @@
+---
+title: "Troubleshooting: Volume cannot be cleaned up after the node of the workload pod is down and recovered"
+author: Derek Su
+draft: false
+date: 2021-11-08
+categories:
+  - "Pod Cleanup"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
+Volume cannot be cleaned up after the node of the workload pod is down and recovered.
+
+## Solution
+
+The root cause is a race condition in the pod cleanup process of Kubernetes.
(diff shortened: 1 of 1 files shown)


content/kb/troubleshooting-volume-cannot-be-cleaned-up-after-the-node-of-the-workload-pod-is-down-and-recovered.md
```diff
@@ -0,0 +1,25 @@
+---
+title: "Troubleshooting: Volume cannot be cleaned up after the node of the workload pod is down and recovered"
```
innobead, line 2:
```suggestion
title: "Troubleshooting: SubPath volume cannot be cleaned up after the node of the workload pod is down and recovered"
```
Is it better to just specify this only for subpath volume?

content/kb/troubleshooting-volume-cannot-be-cleaned-up-after-the-node-of-the-workload-pod-is-down-and-recovered.md
```diff
@@ -0,0 +1,25 @@
...
+  - "Pod Cleanup"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+## Symptoms
+
+Volume cannot be cleaned up after the node of the workload pod is down and recovered.
```
innobead, line 16:
```suggestion
A subPath volume cannot be cleaned up after the node of the workload pod is down and recovered.
```

innobead (changes requested):

content/kb/troubleshooting-volume-cannot-be-cleaned-up-after-the-node-of-the-workload-pod-is-down-and-recovered.md
innobead: @derekbit explained it's general not specific to subpath only.

PR #413, opened 2021-11-10
Create 1.3.0 docs

jenting:

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -14,7 +14,7 @@ style = "paraiso-dark"
 [params]
 description = "Cloud native distributed block storage for Kubernetes"
 tagline = "Easy to use, 100% open source, run anywhere"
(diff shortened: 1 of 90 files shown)


jenting:
I've changed the wording on the upgrade from v1.2.x

innobead (changes requested):
@jenting can't see the 1.3.0 in the preview, probably need to update config.toml

jenting:
> @jenting can't see the 1.3.0 in the preview, probably need to update config.toml

Isn't that mean the 1.3.0 doc will be visible and marked at the latest.

jenting:
Probably we could enhance it by adding the latest version on the config.toml to point what the current latest version. So, it will be like
- 1.3.0
- 1.2.2 (latest)
- 1.2.1
- ...

But we have to make sure using the latest URL redirect to the current version.

innobead:
I think this was misunderstood :) You should make this PR to v1.3.0 instead of the master branch. I just forked the branch.

jenting:
> I think this was misunderstood :) You should make this PR to v1.3.0 instead of the master branch. I just forked the branch.

Understood.

innobead (approved):

PR #417, opened 2021-11-15
Make it clear that some requirements are just for control plan

vietnguyen1254:


content/docs/1.1.0/deploy/install/_index.md
```diff
@@ -46,7 +46,7 @@ For the minimum recommended hardware, refer to the [best practices guide.](../..
 
 ### Using the Environment Check Script
 
-We've written a script to help you gather enough information about the factors.
+We've written a script to help you gather enough information about the factors. NB: Run this only in the control plane node.
```
jenting, line 49: What is NB means?

vietnguyen1254: Ah, it is nota bene meaning note well. We can remove it so people does not confuse :D.

jenting: Would change to "Note" better?

vietnguyen1254: Sure!

innobead:
@vietnguyen1254 Please help update your PR as @jenting review before merging. Also, resolve the DCO errors. Thanks.

content/docs/1.1.0/deploy/install/_index.md
```diff
@@ -21,7 +21,7 @@ For information on deploying Longhorn on specific nodes and rejecting general wo
 Each node in the Kubernetes cluster where Longhorn is installed must fulfill the following requirements:
 
 -  A container runtime compatible with Kubernetes (Docker v1.13+, containerd v1.3.7+, etc.)
--  Kubernetes v1.16+.
+-  Kubernetes v1.16+ (only in control plane node).
     - Recommend Kubernetes v1.17+
```
innobead, line 25: Ideally yes, but I am not certain we can claim this way because of our regression all relying on the same version base including kubelet. 

So it's better just follow up https://kubernetes.io/releases/version-skew-policy/, but here we just claim the min version required.

innobead:
BTW, suggest updating to the latest version instead of an old one (like 1.1.0). Thank you.

PR #422, opened 2021-11-16
Fix recurring job YAML format

c3y1huang:
https://github.com/longhorn/longhorn/issues/3256


innobead (changes requested):
@c3y1huang PR to the master instead of 1.2.2 branch.

PR #430, opened 2021-12-06
Add --pod-selector to the drain instruction

PhanLe1010:
This makes sure that Kubernetes doesn't evict CSI attacher and provisioner so Longhorn can properly detach the volumes

longhorn/longhorn#3304


innobead (approved):
LGTM

PR #439, opened 2021-12-19
Add longhorn-manager webhook server port 9443

jenting:
#### Issue
https://github.com/longhorn/longhorn/issues/3513

### DIFF
--- a/content/docs/1.3.0/references/networking.md
+++ b/content/docs/1.3.0/references/networking.md
@@ -33,6 +33,12 @@ To | Port | Protocol
 `External Backupstore` | User defined | TCP
 `Kubernetes API server` | `Kubernetes API server port` | TCP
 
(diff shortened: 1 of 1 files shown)


content/docs/1.3.0/references/networking.md
```diff
@@ -33,6 +33,17 @@ To | Port | Protocol
...
+### Longhorn Webhook
+#### Ingress:
+From | Port | Protocol
+--- | --- | ---
+`Kubernetes API server` | 9443 | TCP
+
+#### Egress:
+To | Port | Protocol
+--- | --- | ---
+`Kubernetes API server` | 9443 | TCP
```
innobead, line 45: Do we need egress?

innobead (changes requested):

innobead (approved):

PR #440, opened 2021-12-20
Add a doc that instructs how users can restore Longhorn to a new cluster 

shuo-wu:
https://github.com/longhorn/longhorn/issues/3367


jenting (commented):
CI failed

innobead (approved):

PR #443, opened 2022-01-05
Add KB: Troubleshooting: Upgrading volume engine is stuck in deadlock

PhanLe1010:
longhorn/longhorn#2697

### DIFF
--- a/content/kb/troubleshooting-engine-upgrading-stuck-in-deadlock.md
+++ b/content/kb/troubleshooting-engine-upgrading-stuck-in-deadlock.md
@@ -1,0 +1,81 @@
+---
+title: "Troubleshooting: Upgrading volume engine is stuck in deadlock"
+author: Phan Le
+draft: false
+date: 2022-01-03
+categories:
+- "Longhorn Upgrade"
+---
+
+## Applicable versions
+
+This happens when users upgrade Longhorn from version <= v1.1.1 to a newer version.
+
+## Symptoms
+
+[Upgrading Longhorn system](https://longhorn.io/docs/1.2.3/deploy/upgrade/) includes 2 steps: first upgrade Longhorn manager to the latest version,
+then upgrade the Longhorn engine to the latest version using the latest Longhorn manager.
+When doing the second step (upgrading Longhorn engine), you may hit the problem that some volumes are stuck in engine upgrading.
+You may also see that volume attachment/detachment cannot finish (e.g., Longhorn volumes are stuck in detaching or attaching state).
+
+## Reason
+
+There is a bug Longhorn version <= v1.1.1 which leads to a deadlock in the instance manager pods.
+See more details at https://github.com/longhorn/longhorn/issues/2697.
+When you upgrade Longhorn from version <= v1.1.1 to a newer version, you may hit this bug in a cluster with a few hundred volumes.
+
+## Solution
+
+### Prevent the deadlock from happening
+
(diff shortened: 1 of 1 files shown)


content/kb/troubleshooting-engine-upgrading-stuck-in-deadlock.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Prevent the deadlock from happening
+
+We fixed this bug in Longhorn version >= v1.1.2.
+If you are planning to upgrade Longhorn to a version >= v1.1.2, you can follow the following steps to avoid the bug:
+
+1. Upgrade Longhorn manager as normal. See https://longhorn.io/docs/1.2.3/deploy/upgrade/longhorn-manager/
+1. At this moment, there will be 2 different instance-manager versions running side by side in your cluster (1 pod for the old version and 1 new pod for the new version).
+   The engine/replica processes will continue to live inside the old instance-manager pods.
+   Now, if for some reason the volumes are detached (because you scale down the workload or because you roll out the workload),
```
joshimoo, line 37: the user needs to ensure that the volume goes through a attach/detach cycle. This might not happen in the case of a rollout where there are multiple pods on Node X that are all using Volume Y since there is configuration for maxSurge, maxUnavailable etc. https://www.educative.io/blog/kubernetes-deployments-strategies

Multiple pods all using an RWO volume is fine as long as their are on the same node since RWO/RWX AccessMode is based on Node.

The easiest solution to ensure, correct scale down & up is to do a with an **appropriate filter (example, only drain the workload pods via some kubectl & bash magic)** `kubectl drain --pod-selector='!longhorn.io/component,app!=csi-attacher,app!=csi-provisioner' <NODE> --ignore-daemonsets`

content/kb/troubleshooting-engine-upgrading-stuck-in-deadlock.md
```diff
@@ -0,0 +1,77 @@
...
+        mv grpcurl /usr/local/bin/
+        # Call instance manager gRPC APIs
+        wget https://raw.githubusercontent.com/longhorn/longhorn-instance-manager/master/pkg/rpc/rpc.proto
+        wget https://raw.githubusercontent.com/grpc/grpc/master/src/proto/grpc/health/v1/health.proto
+        # check the health of grpc server on the instance manager instance-manager-e-f386c595
+        grpcurl -d '' -plaintext -import-path ./ -proto health.proto <INSTANCE-MANAGER-IP>:8500 grpc.health.v1.Health/Check
+        # Server returns "status": "SERVING"
+        grpcurl -d '' -plaintext -import-path ./ -proto rpc.proto <INSTANCE-MANAGER-IP>:8500 ProcessManagerService/ProcessList
+        # If the server never returns response, this is a stuck instance manager pod
+        ```
```
joshimoo, line 67: This part is to verify that the IM has deadlocked

joshimoo (approved):
LGTM, can be merged, 
Would be nice to add the clarification about the rollout + drain example.

content/kb/troubleshooting-engine-upgrading-stuck-in-deadlock.md
PhanLe1010: Thank you! Updated the doc with drain command

innobead:
cc @longhorn/qa for knowledge sharing

PR #458, opened 2022-02-23
Add detailed list of which directories the longhorn components need

jenting:
https://github.com/longhorn/longhorn/issues/3549

### DIFF
--- a/content/docs/1.2.3/deploy/install/_index.md
+++ b/content/docs/1.2.3/deploy/install/_index.md
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
 
 ### Root and Privileged Permission
 
-The longhorn manager uses `Bidirectional` mount propagation to access the host's disk (default `/var/lib/longhorn`) in the container.
-An use case is to access disk metadata and replicas directory on the host.
+Longhorn components require root access with privileged permissions to achieve volume operations and management, because Longhorn relies on system resources on the host across different namespaces, for example, Longhorn uses `nsenter` to understand block devices' usage or encrypt/decrypt volumes on the host.
 
-Besides, Longhorn uses `nsenter` command to execute commands on the host. The `nsenter` command, for example, is to check the host's disk space usage
-```shell
-stat /var/lib/longhorn/ -fc '{"path":"%n","fsid":"%i","type":"%T","freeBlock":%f,"totalBlock":%b,"blockSize":%s}'
-```
-The `nsenter` command rely on privileged containers and `hostPID: true`.
+Below are the directories Longhorn components requiring access with root and privileged permissions :
+
+- Longhorn Manager
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` to access the mounts on the host to understand disks usage.
+  - /var/lib/longhorn: The default path for storing volume data on a host.
+- Longhorn Engine Image
+  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
+- Longhorn Instance Manager (Engine)
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` to manage iSCSI targets and initiators, also some file system operations like sync before snapshotting.
+  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
+- Longhorn Instance Manager (Replica)
+  - /: Access any data path on this node and access Longhorn engine binaries.
+- Longhorn Share Manager
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /lib/modules: Kernel modules requried by `cryptsetup` for volume encryption.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` for volume encryption.
+  - /sys: Support volume encryption by `cryptsetup`.
+- Longhorn CSI Plugin
+  - /: For host checks via the NFS customer mounter (deprecated). Note that, this will be removed in the future release.
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /lib/modules: Kernel modules required by Longhorn CSI plugin.
+  - /sys: Support volume encryption by `cryptsetup`.
+  - /var/lib/kubelet/plugins/kubernetes.io/csi: The path where the Longhorn CSI plugin creates the staging path (via `NodeStageVolume`) of a block device. The staging path will be bind-mounted to the target path `/var/lib/kubelet/pods` (via `NodePublishVolume`) for support single volume could be mounted to multiple Pods.
+  - /var/lib/kubelet/plugins_registry: The path where the node-driver-registrar registers the CSI plugin with kubelet.
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
+  - /var/lib/kubelet/pods: The path where the Longhorn CSI driver mounts volume from the target path (via `NodePublishVolume`).
+- Longhorn CSI Attacher/Provisioner/Resizer/Snapshotter
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
+- Longhorn Backing Image Manager
+  - /var/lib/longhorn: The default path for storing data on the host.
+- Longhorn Backing Image Data Source
+  - /var/lib/longhorn: The default path for storing data on the host.
 
 ### Installing open-iscsi
 
--- a/content/docs/1.2.4/deploy/install/_index.md
+++ b/content/docs/1.2.4/deploy/install/_index.md
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
 
 ### Root and Privileged Permission
 
-The longhorn manager uses `Bidirectional` mount propagation to access the host's disk (default `/var/lib/longhorn`) in the container.
-An use case is to access disk metadata and replicas directory on the host.
+Longhorn components require root access with privileged permissions to achieve volume operations and management, because Longhorn relies on system resources on the host across different namespaces, for example, Longhorn uses `nsenter` to understand block devices' usage or encrypt/decrypt volumes on the host.
 
-Besides, Longhorn uses `nsenter` command to execute commands on the host. The `nsenter` command, for example, is to check the host's disk space usage
-```shell
-stat /var/lib/longhorn/ -fc '{"path":"%n","fsid":"%i","type":"%T","freeBlock":%f,"totalBlock":%b,"blockSize":%s}'
-```
-The `nsenter` command rely on privileged containers and `hostPID: true`.
+Below are the directories Longhorn components requiring access with root and privileged permissions :
+
+- Longhorn Manager
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` to access the mounts on the host to understand disks usage.
+  - /var/lib/longhorn: The default path for storing volume data on a host.
+- Longhorn Engine Image
+  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
+- Longhorn Instance Manager (Engine)
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` to manage iSCSI targets and initiators, also some file system operations like sync before snapshotting.
+  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
+- Longhorn Instance Manager (Replica)
+  - /: Access any data path on this node and access Longhorn engine binaries.
+- Longhorn Share Manager
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /lib/modules: Kernel modules requried by `cryptsetup` for volume encryption.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` for volume encryption.
+  - /sys: Support volume encryption by `cryptsetup`.
+- Longhorn CSI Plugin
+  - /: For host checks via the NFS customer mounter (deprecated). Note that, this will be removed in the future release.
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /lib/modules: Kernel modules required by Longhorn CSI plugin.
+  - /sys: Support volume encryption by `cryptsetup`.
+  - /var/lib/kubelet/plugins/kubernetes.io/csi: The path where the Longhorn CSI plugin creates the staging path (via `NodeStageVolume`) of a block device. The staging path will be bind-mounted to the target path `/var/lib/kubelet/pods` (via `NodePublishVolume`) for support single volume could be mounted to multiple Pods.
+  - /var/lib/kubelet/plugins_registry: The path where the node-driver-registrar registers the CSI plugin with kubelet.
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
+  - /var/lib/kubelet/pods: The path where the Longhorn CSI driver mounts volume from the target path (via `NodePublishVolume`).
+- Longhorn CSI Attacher/Provisioner/Resizer/Snapshotter
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
+- Longhorn Backing Image Manager
+  - /var/lib/longhorn: The default path for storing data on the host.
+- Longhorn Backing Image Data Source
+  - /var/lib/longhorn: The default path for storing data on the host.
 
 ### Installing open-iscsi
 


content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,8 +86,41 @@ If MountPropagation is disabled, Base Image feature will be disabled.
 
 ### Root and Privileged Permission
 
-The longhorn manager uses `Bidirectional` mount propagation to access the host's disk (default `/var/lib/longhorn`) in the container.
-An use case is to access disk metadata and replicas directory on the host.
+The longhorn components need to access the host's disk in the container.
```
innobead, line 89:
```suggestion
Longhorn components require root access with privileged permissions to achieve volume operations and management, because Longhorn relies on system resources management on the host across different namespaces, for example, Longhorn uses `nsenter` to understand block devices' usage or encrypt/decrypt volumes on the host.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,8 +86,41 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+A use case is to access disk metadata and replicas directory on the host. The list of the directories of all components are:
```
innobead, line 90: Below are examples to list the different use cases requiring root access and privileged permissions.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,8 +86,41 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+
+- Longhorn Manager
+  - /dev: Volume block device
```
innobead, line 93: Suggest having an explanatory brief statement to say what we need to access `/dev` instead of just a noun. The same as others.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,8 +86,41 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/kubelet/plugins/driver.longhorn.io: Kubernetes CSI plugin
+  - /var/lib/kubelet/pods: Kubernetes CSI plugin
+- Longhorn CSI Attacher/Provisioner/Resizer/Snapshotter
+  - /var/lib/kubelet/plugins/driver.longhorn.io: Kubernetes CSI plugin
+- Longhorn Backing Image Manager
+  - /var/lib/longhorn: Data path
+- Longhorn Backing Image Data Source
+  - /var/lib/longhorn: Data path
 
 Besides, Longhorn uses `nsenter` command to execute commands on the host. The `nsenter` command, for example, is to check the host's disk space usage
```
innobead, line 125: We can remove this part, but move to the summary instead.

innobead (changes requested):

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/longhorn/engine-binaries: Longhorn engine binary
+- Longhorn Instance Manager (Replica)
+  - /: Data path (allow the user to customize data path)
+- Longhorn Share Manager
+  - /dev: Volume block device
+  - /lib/modules: volume encryption by `cryptsetup`
+  - /proc: To access host's mount namespace by `nsenter`
+  - /sys: Volume encryption by `cryptsetup`
+- Longhorn CSI Plugin
+  - /: For RWX shared volume
```
shuo-wu, line 111: For this volume mount, here is the more detailed explanation: https://github.com/longhorn/longhorn-manager/blob/8177316cafe002b512577bb4f05a2065e80ed383/csi/node_server.go#L610

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /dev: Volume block device
+  - /lib/modules: Kubernetes CSI plugin
+  - /sys: Volume encryption by `cryptsetup`
+  - /var/lib/kubelet/plugins/kubernetes.io/csi: Kubernetes CSI plugin
+  - /var/lib/kubelet/plugins_registry: Kubernetes CSI plugin
+  - /var/lib/kubelet/plugins/driver.longhorn.io: Kubernetes CSI plugin
+  - /var/lib/kubelet/pods: Kubernetes CSI plugin
```
shuo-wu, line 118: Not sure if we need to explain this part a little bit more.

shuo-wu (changes requested):

content/docs/1.2.3/deploy/install/_index.md
jenting: I thought the reason is we need to support RWX volume https://github.com/longhorn/longhorn-manager/commit/a644b94439754eb791cf31955ec8b8bff8abf30b.

But if we want to grant permission to `nsenter`, shouldn't we just need to allow `/proc` mount point rather than the root mount point?

content/docs/1.2.3/deploy/install/_index.md
innobead: Probably we should.  The intention is to let user understand why the CSI compatible storage solution like longhorn need these privileged permissions besides our components. However, the general explanation is good enough.

content/docs/1.2.3/deploy/install/_index.md
shuo-wu: Yeah, I am confused about this as well. The mounter is actually [this script](https://github.com/longhorn/longhorn-manager/blob/master/package/nsmounter), in which we use `-t 1` rather than something like `--net=/host/proc/1/ns/net --mount=/host/proc/1/ns/mnt`. @joshimoo Can you provide more details?

joshimoo: @jenting @shuo-wu We no longer need the `/rootfs` since I have removed the nfs mounter and replaced it with the nsenter script above. Since we run the csi plugin with `hostPid: true` we can utilize `-t 1` since `/proc` maps to `/proc` in this case.

It's a left over from before where we were doing host checks via the custom mounter.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/longhorn/engine-binaries: The path where the Longhorn engine binaries located.
+- Longhorn Instance Manager (Replica)
+  - /: Access any data path on this node and access Longhorn engine binaries.
+- Longhorn Share Manager
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
+  - /lib/modules: Kernel module requried by `cryptsetup`.
+  - /proc: To access host's mount namespace by `nsenter` across different namepsaces.
+  - /sys: To support volume encryption by `cryptsetup`.
+- Longhorn CSI Plugin
+  - /: For RWX shared volume.
```
joshimoo, line 111: "/" should no longer be needed with the new nsenter shell script and the removed nfsmounter.
Since we are running with `hostPid: true` -> `/proc => /proc`

content/docs/1.2.3/deploy/install/_index.md
jenting: @joshimoo So we could remove these lines https://github.com/longhorn/longhorn-manager/blob/1664880ddaedb9875671714b001a2ad7c9bc3b5c/csi/deployment.go#L525-L532 and https://github.com/longhorn/longhorn-manager/blob/1664880ddaedb9875671714b001a2ad7c9bc3b5c/csi/deployment.go#L459-L463, right?

content/docs/1.2.3/deploy/install/_index.md
jenting: Since within v1.2.3 Longhorn code, we still have mount point root mount point inside the CSI plugins.
Therefore, we should list the root mount point in the doc to let the user/customer knows the purpose from the security point of view.

content/docs/1.2.3/deploy/install/_index.md
shuo-wu: Then we can mention that this will be removed since v1.2.4.

jenting: I'd prefer to add the wording that this will be removed in the coming release. 

I'm not sure if we can get this [improvement](https://github.com/longhorn/longhorn/issues/3736) into v1.2.4.
cc @innobead

innobead: I don't think we should at this moment if it's not in 1.2.4 yet.

shuo-wu: Sure. Then we don't need to mention in which specific version we will remove it.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Share Manager
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
+  - /lib/modules: Kernel module requried by `cryptsetup`.
+  - /proc: To access host's mount namespace by `nsenter` across different namepsaces.
+  - /sys: To support volume encryption by `cryptsetup`.
+- Longhorn CSI Plugin
+  - /: For host checks via the NFS customer mounter (deprecated). Note that, this will be removed in the future release.
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
+  - /lib/modules: Kernel module required by Kubernetes CSI plugin.
+  - /sys: To support volume encryption by `cryptsetup`.
```
shuo-wu, line 114: For the CSI plugin, this mount is introduced before implementing the volume encryption. Is there any other reason for it?

shuo-wu: I am not familiar with it, either. Not sure if others know anything about this.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
-An use case is to access disk metadata and replicas directory on the host.
-
-Besides, Longhorn uses `nsenter` command to execute commands on the host. The `nsenter` command, for example, is to check the host's disk space usage
-```shell
-stat /var/lib/longhorn/ -fc '{"path":"%n","fsid":"%i","type":"%T","freeBlock":%f,"totalBlock":%b,"blockSize":%s}'
-```
-The `nsenter` command rely on privileged containers and `hostPID: true`.
+Longhorn components require root access with privileged permissions to achieve volume operations and management, because Longhorn relies on system resources on the host across different namespaces, for example, Longhorn uses `nsenter` to understand block devices' usage or encrypt/decrypt volumes on the host.
+
+Below are the list of the directories of all components are:
```
innobead, line 91:
```suggestion
Below are the directories Longhorn components requiring access with root and privileged permissions :
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+
+- Longhorn Manager
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
```
innobead, line 94:
```suggestion
  - /dev: Block devices created by Longhorn are under the `/dev` path.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /proc: To access host's mount namespace by `nsenter` across different namepsaces.
```
innobead, line 95:
```suggestion
  - /proc: Find the recognized host process like container runtime, then use `nsenter` to access the mounts on the host to understand disks usage.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/longhorn: The path to use for storing data on a host.
```
innobead, line 96:
```suggestion
  - /var/lib/longhorn: The default path for storing volume data on a host.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Engine
+  - /var/lib/longhorn/engine-binaries: The path where the Longhorn engine binaries located.
```
innobead, line 98:
```suggestion
  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Instance Manager (Engine)
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
```
innobead, line 100:
```suggestion
  - /dev: Block devices created by Longhorn are under the `/dev` path.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /proc: To access host's mount namespace by `nsenter` across different namepsaces.
```
innobead, line 101:
```suggestion
  - /proc: Find the recognized host process like container runtime, then use `nsenter` to manage iSCSI targets and initiators, also some file system operations like sync before snapshotting. 
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/longhorn/engine-binaries: The path where the Longhorn engine binaries located.
```
innobead, line 102:
```suggestion
  - /var/lib/longhorn/engine-binaries: The default path for storing the Longhorn engine binaries.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Instance Manager (Replica)
+  - /: Access any data path on this node and access Longhorn engine binaries.
```
innobead, line 104: Saw it requires the root path, but what cases actually happening in IM-R pods use that path?

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Share Manager
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
```
innobead, line 106:
```suggestion
  - /dev: Block devices created by Longhorn are under the `/dev` path.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /lib/modules: Kernel module requried by `cryptsetup`.
```
innobead, line 107:
```suggestion
  - /lib/modules: Kernel modules requried by `cryptsetup` for volume encryption.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /proc: To access host's mount namespace by `nsenter` across different namepsaces.
```
innobead, line 108:
```suggestion
  - /proc: Find the recognized host process like container runtime, then use `nsenter` for volume encryption.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /sys: To support volume encryption by `cryptsetup`.
```
innobead, line 109:
```suggestion
  - /sys: Support volume encryption by `cryptsetup`.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn CSI Plugin
+  - /: For host checks via the NFS customer mounter (deprecated). Note that, this will be removed in the future release.
+  - /dev: Longhorn is a block storage solution for Kubernetes. The block device is created under `/dev` path.
```
innobead, line 112:
```suggestion
  - /dev: Block devices created by Longhorn are under the `/dev` path.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /lib/modules: Kernel module required by Kubernetes CSI plugin.
```
innobead, line 113:
```suggestion
  - /lib/modules: Kernel modules required by Longhorn CSI plugin.
```

content/docs/1.2.3/deploy/install/_index.md
innobead, line 114:
```suggestion
  - /sys: Support volume encryption by `cryptsetup`.
```

innobead: yeah, we can figure out it later. 

cc @derekbit @joshimoo

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /sys: To support volume encryption by `cryptsetup`.
+  - /var/lib/kubelet/plugins/kubernetes.io/csi: The path where the Kubernetes CSI plugin creates a staging path (via `NodeStageVolume`) of a block device. Then, Kubernetes CSI plugin bind mount the staging path to target path `/var/lib/kubelet/pods` (via `NodePublishVolume`) for support single volume could be mounted to multiple Pods.
```
innobead, line 115:
```suggestion
  - /var/lib/kubelet/plugins/kubernetes.io/csi: The path where the Longhorn CSI plugin creates the staging path (via `NodeStageVolume`) of a block device. The staging path will be bind-mounted to the target path `/var/lib/kubelet/pods` (via `NodePublishVolume`) for support single volume could be mounted to multiple Pods.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/kubelet/plugins_registry: The path where the node-driver-registrar registers the CSI plugin with kubelet.
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for kubelet -> CSI driver communication is done.
```
innobead, line 117:
```suggestion
  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+  - /var/lib/kubelet/pods: The path where the CSI driver mounts volume from staging to target path (via `NodePublishVolume`).
```
innobead, line 118:
```suggestion
  - /var/lib/kubelet/pods: The path where the Longhorn CSI driver mounts volume from the target path (via `NodePublishVolume`).
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn CSI Attacher/Provisioner/Resizer/Snapshotter
+  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for kubelet -> CSI driver communication is done.
```
innobead, line 120:
```suggestion
  - /var/lib/kubelet/plugins/driver.longhorn.io: The path where the socket for the communication between kubelet Longhorn CSI driver.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Backing Image Manager
+  - /var/lib/longhorn (the default data path, could be customized by the user): The path to use for storing data on a host.
```
innobead, line 122:
```suggestion
  - /var/lib/longhorn: The default path for storing data on the host.
```

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
+- Longhorn Backing Image Data Source
+  - /var/lib/longhorn (the default data path, could be customized by the user): The path to use for storing data on a host.
```
innobead, line 124:
```suggestion
  - /var/lib/longhorn: The default path for storing data on the host.
```

innobead (changes requested):
LGTM, just a few comments needed to resolve, and some wording needed to change.

content/docs/1.2.3/deploy/install/_index.md
jenting: I think it simplifies the customize path for storing volume data on a host.

Below is the command to launch a replica process with the specific path to storing the replica data.
```shell
/host/var/lib/longhorn/engine-binaries/rancher-mirrored-longhornio-longhorn-engine-v1.1.1/longhorn replica /host/var/lib/longhorn/replicas/pvc-3a27bbc4-bd70-4a93-8aca-154ff88df514-9ead319e --size 1073741824 --listen 0.0.0.0:10000
```

But the better approach _probably_ mounts a different hostpath if the users customize the longhorn storage path rather than the root mount point.

content/docs/1.2.3/deploy/install/_index.md
```diff
@@ -86,14 +86,42 @@ If MountPropagation is disabled, Base Image feature will be disabled.
...
-The `nsenter` command rely on privileged containers and `hostPID: true`.
+Longhorn components require root access with privileged permissions to achieve volume operations and management, because Longhorn relies on system resources on the host across different namespaces, for example, Longhorn uses `nsenter` to understand block devices' usage or encrypt/decrypt volumes on the host.
+
+Below are the directories Longhorn components requiring access with root and privileged permissions :
+
+- Longhorn Manager
+  - /dev: Block devices created by Longhorn are under the `/dev` path.
+  - /proc: Find the recognized host process like container runtime, then use `nsenter` to access the mounts on the host to understand disks usage.
+  - /var/lib/longhorn: The default path for storing volume data on a host.
+- Longhorn Engine
```
innobead, line 97: Do we need this? I believe this should be covered in IM engine instead.

innobead (changes requested):

content/docs/1.2.3/deploy/install/_index.md
jenting: This should be Longhorn Engine Image DaemonSet, I'll rename it.

innobead (approved):

innobead:
Thanks @jenting 👍

PR #460, opened 2022-02-24
Add note in upgrade doc for two different Rancher UI

jenting:


content/docs/1.2.3/deploy/upgrade/longhorn-manager.md
```diff
@@ -29,12 +29,26 @@ If Longhorn was installed using a Helm Chart, or if it was installed as Rancher
 
 > **Prerequisite:** Always back up volumes before upgrading. If anything goes wrong, you can restore the volume using the backup.
 
+#### Upgrade as a Rancher Catalog App
+
+To upgrade Longhorn App, make sure which Rancher UI the existing Longhorn App was installed with. There are two Rancher UIs, one is Cluster Manager (old UI), the other one is Cluster Explorer (new UI). The Longhorn App in different UIs considered as two different applications by Rancher. They cannot upgrade to each other. If you installed Longhorn in Cluster Manager (old UI) before, you need to use Cluster Manager (old UI) to upgrade Longhorn to a newer version. And vice versa.
```
innobead, line 34:
```suggestion
To upgrade the Longhorn App, make sure which Rancher UI the existing Longhorn App was installed with. There are two Rancher UIs, one is Cluster Manager (old UI), and the other one is Cluster Explorer (new UI). The Longhorn App in different UIs considered as two different applications by Rancher. They cannot upgrade to each other. If you installed Longhorn in the Cluster Manager, you need to use the Cluster Manager to upgrade Longhorn to a newer version, and vice versa for the Cluster Explorer.
```

innobead (changes requested):

innobead (approved):

PR #487, opened 2022-04-12
Improve the doc for volume actual size and space usage

derekbit:
[Longhorn 3627](https://github.com/longhorn/longhorn/issues/3627)


content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,69 @@ In this section, you'll have a better understanding of concepts related to volum
...
+
+2. Do not retain too many snapshots for the volumes.
+
+3. Clean up snapshots will help reclaim disk space. There are two ways to clean up snapshots:
+    - Delete the snapshots manually via Longhorn UI.
+    - Set a snapshot recurring job with retention 1, then the snapshots will be cleaned up automatically.
+ 
+    Also, notice that the extra space, up to volume nominal `size`, is required during snapshot cleanup and merge.
+
+4. Set `Storage Over Provisioning Percentage` to be smaller than 100% if there are write-heavy volumes.
```
shuo-wu, line 86: Maybe you can combine 1st item with 4th item.
- Reserve enough free space in disks as buffers in case of the actual size of existing volumes keep growing up. 
  - A quick estimation for maximum space consumption estimation of a volume is: 
    ```
    (N + 1 + 1) x head/snapshot average actual size
    ```
    where N is the number of snapshots retained, and the 1st 1 is for the system snapshot, and 2nd 1 is for the tmp space may required by snapshot deletion. 
  - If there are heavy writing tasks for volumes, `head/snapshot average actual size` would be volume the nominal size. And in this case, it's better to set `Storage Over Provisioning Percentage` to be smaller than 100%
to avoid disk space exhaustion.

shuo-wu (commented):
Actually, I just joined a call with UiPath team. One of the questions they wanted to understand was exactly the filesystem part in this doc.

Based on the discussion, I think we can:
1. Add more illustrations for this part.
2. Add one more suggestion with the following example based on this behavior. Assume users write 50Gi data to the filesystem of a volume, delete it, then re-write it. If the volume spec size is 100Gi, the actual size at the end would range from 50Gi - 100Gi since the filesystem may or maynot reuse the dirty blocks. But if the volume spec size is 60Gi, the actual size at the end would range from 50Gi - 60Gi. Here the filesystem has to reuse the dirty blocks in the 2nd round of writing (even if the filesystem itself is unaware of the reusage). Allocating an appropriate spec/nominal size for a volume that holds heavy writing tasks would make disk space usage more efficient.

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,74 @@ In this section, you'll have a better understanding of concepts related to volum
...
+
+
+4. Then, rewrite the 50Gi data, and the `df` command in the filesystem shows 50Gi used space again. Howerver, the `actual size` is increased by 50Gi and becomes 100Gi.
+
+    - After deletion, filesystem may or maynot reuse the recently freed blocks from recent deleted files according to the filesystem design and please refer to [Block allocation strategies of various filesystems](https://www.ogris.de/blkalloc). If the volume nominal `size` is 100Gi, the `actual size` in the end would range from 50Gi to 100Gi since the filesystem may or maynot reuse the freed blocks. On the other hand, if the volume nominal `size` is 60Gi, the `actual size` at the end would range from 50Gi to 60Gi, because the filesystem has to reuse the freed blocks in the 2nd round of writing. Thus, allocating an appropriate nominal `size` for a volume that holds heavy writing tasks according to the IO pattern would make disk space usage more efficient.
+
+### An Example That Helps Understand Volume `Size` and `Actual Size`:
```
shuo-wu, line 42: There are overlapping parts between the 2 examples. I think we can combine them and emphasize the following key points (with more illustrations):
1. Dirty block not releasing issue: Longhorn does not support TRIM/UNMAP operations. Hence deleting files from filesystems won't lead to volume actual size decreasing/shrinking. 
2. Dirty block not reusing issue: Deleting then writing new files would lead to the actual size keeps increasing. Since the filesystems may not reuse the dirty blocks (the recently freed blocks from recent deleted files). Thus, allocating an appropriate nominal `size` for a volume that holds heavy writing tasks according to the IO pattern would make disk space usage more efficient. 
3. Deleting snapshots **may** (not guarantee) eliminate the overlapping part of the used blocks (no matter the blocks are recently released blocks by fs or still hold historical data).

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,74 @@ In this section, you'll have a better understanding of concepts related to volum
...
+
 {{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-3.png" >}}
 5. Delete snapshot#1 and wait for snapshot purge complete. See Figure 4 of the illustration.
     - Here Longhorn actually coalesces the snapshot#1 with the snapshot#2.
     - For the overlapping part during the coalesce, the newer data (data#2) will be retained in the blocks. Then some historical data is removed and the volume gets shrunk (from 6.1Gi to 4.65Gi in the example).
 {{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-4.png" >}}
+
+
+
+## Space Configuration Suggestions for Volumes
```
shuo-wu, line 74: Add suggestion:
4. An appropriate the spec/nominal size

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,81 @@ In this section, you'll have a better understanding of concepts related to volum
...
+
+{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-illustration-fig5.png" >}}
+
+8. Delete snapshot#1 and wait for snapshot purge complete. See Figure 7 of the illustration.
     - Here Longhorn actually coalesces the snapshot#1 with the snapshot#2.
-    - For the overlapping part during the coalesce, the newer data (data#2) will be retained in the blocks. Then some historical data is removed and the volume gets shrunk (from 6.1Gi to 4.65Gi in the example).
-{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-4.png" >}}
+    - For the overlapping part during the coalesce, the newer data (data#2) will be retained in the blocks. Then some historical data is removed and the volume gets shrunk (from 16.3 Gi to 11.4 Gi in the example).
+
+{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-illustration-fig6.png" >}}
```
shuo-wu, line 78: NIT: We can add a brief summary at the end, which includes the 3 key points I mentioned last time.

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,81 @@ In this section, you'll have a better understanding of concepts related to volum
...
 
-1. Create a 5Gi volume, then attach and mount it on a node. See Figure 1 of the illustration.
-    - For this empty volume, the nominal `size` is 5Gi and the `actual size` is almost 0.
-    - There is some meta info in the volume hence the `actual size` is not exactly 0.
-2. Write 2Gi data (data#1) in the volume mount point and take a snapshot (snapshot#1). See Figure 2 of the illustration.
-    - Now data#1 is stored in snapshot#1 and the `actual size` is 2Gi.
-{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-1.png" >}}
-3. Delete data#1 from the mount point.
-    - The truth of the data#1 deletion is that the data#1 is marked as removed in **the filesystem level** (For example, inode deletion in ext4). Since **Longhorn operates on the block level** and does not understand the filesystem, as a result, the disk blocks/space storing data#1 won't be released after the deletion.
+{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-illustration.png" >}}
```
shuo-wu, line 28: NIT: You can replace the description of the gray part with `the blocked freed by the file deletion of filesystem but still considered as allocated in Longhorn's perspective`

shuo-wu (approved):
In general LGTM

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+
+8. Delete snapshot#1 and wait for snapshot purge complete. See Figure 7 of the illustration.
     - Here Longhorn actually coalesces the snapshot#1 with the snapshot#2.
-    - For the overlapping part during the coalesce, the newer data (data#2) will be retained in the blocks. Then some historical data is removed and the volume gets shrunk (from 6.1Gi to 4.65Gi in the example).
-{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-4.png" >}}
+    - For the overlapping part during the coalesce, the newer data (data#2) will be retained in the blocks. Then some historical data is removed and the volume gets shrunk (from 16.3 Gi to 11.4 Gi in the example).
+
+{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-illustration-fig6.png" >}}
+
+Here we summarize the issues we encountered in the example:
```
innobead, line 82:
```suggestion
Here we summarize the important things related to disk space usage we have in the example:
```

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+1. Unused blocks not releasing issue
```
innobead, line 83:
```suggestion
1. Unused blocks are not released
```

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+    Longhorn does not support TRIM/UNMAP operations. Hence deleting files from filesystems will not lead to volume actual size decreasing/shrinking.
```
innobead, line 84: Add an extra new line for formatting, otherwise, this will just immediately follow up with the above line.

```suggestion

    Longhorn does not support TRIM/UNMAP operations. Hence deleting files from filesystems will not lead to volume actual size decreasing/shrinking.
```

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+2. Allocated but unused blocks not reusing issue
```
innobead, line 85:
```suggestion
2. Allocated blocks but unused are not reused
```

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+    Deleting then writing new files would lead to the actual size keeps increasing. Since the filesystem may not reuse the recently freed blocks from recently deleted files. Thus, allocating an appropriate nominal size for a volume that holds heavy writing tasks according to the IO pattern would make disk space usage more efficient.
```
innobead, line 86:
```suggestion

    Deleting then writing new files would lead to the actual size keeps increasing. Since the filesystem may not reuse the recently freed blocks from recently deleted files. Thus, allocating an appropriate nominal size for a volume that holds heavy writing tasks according to the IO pattern would make disk space usage more efficient.
```

innobead (changes requested):

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
-4. Write 4Gi data (data#2) in the volume mount, then take one more snapshot (snapshot#2). See Figure 3 of the illustration.
-    - Now the `actual size` is 6Gi, which is greater than the nominal `size`.
-    - There is overlapping between 2 snapshots in block level (See the 2 snapshots in Figure 3), since the data#1 is marked as removed in snapshot#2 then the filesystem reuses the space.
-{{< figure src="/img/screenshots/volumes-and-nodes/volume-size-example-3.png" >}}
-5. Delete snapshot#1 and wait for snapshot purge complete. See Figure 4 of the illustration.
+    - The `actual size` is still 8.25 Gi.
+
+7. Write 8 Gi data (data#2) in the volume mount, then take one more snapshot (snapshot#2). See Figure 5 of the illustration.
+    - Now the `actual size` is 16.3 Gi, which is greater than the volume nominal `size`.
+    - There is overlapping between 2 snapshots in block level (See the 2 snapshots in Figure 6), since the data#1 is marked as removed in snapshot#2 then the filesystem reuses the space.
```
shuo-wu, line 68: From a filesystem's perspective, the overlapping part (between the 2 snapshots) is considered as the blocks that have to be reused or overwritten. But in terms of Longhorn, these blocks are actually fresh ones held in another snapshot/volume head. See the 2 snapshots in Figure 6.

shuo-wu (approved):
LGTM

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+## Space Configuration Suggestions for Volumes
+
+1. Reserve enough free space in disks as buffers in case of the actual size of existing volumes keep growing up.
+    - A quick estimation for maximum space consumption estimation of a volume is
+  
+        ```
+        (N + 1 + 1) x head/snapshot average actual size
+        ```
+
+        where N is the number of snapshots retained, and the 1st 1 is for the system snapshot, and 2nd 1 is for the tmp space may required by snapshot deletion.
```
innobead, line 97:
```suggestion
        where N is the number of snapshots retained, the 1st 1 is for the system snapshot, and the 2nd 1 is for the temporary space that may be required by snapshot deletion.
```

content/docs/1.2.4/volumes-and-nodes/volume-size.md
```diff
@@ -19,29 +19,90 @@ In this section, you'll have a better understanding of concepts related to volum
...
+    - If there are heavy writing tasks for volumes, head/snapshot average actual size would be volume the nominal size. In this case, it's better to set `Storage Over Provisioning Percentage` to be smaller than 100% to avoid disk space exhaustion.
```
innobead, line 98:
```suggestion
    - If there are heavy writing tasks for volumes, the head/snapshot average actual size would be volume the nominal size. In this case, it's better to set `Storage Over Provisioning Percentage` to be smaller than 100% to avoid disk space exhaustion.
```

innobead (changes requested):

innobead:
LGTM, this is pretty good. After the suggested changes to resolve, then we can merge this.

innobead (approved):

PR #494, opened 2022-04-29
Add orphaned-data-cleanup doc

derekbit:
https://github.com/longhorn/longhorn/issues/685
https://github.com/longhorn/longhorn/issues/3853


innobead (approved):
LGTM

PR #512, opened 2022-06-07
cluster-autoscaler: add document

c3y1huang:
https://github.com/longhorn/longhorn/issues/3884


derekbit (commented):
LGTM.
Do we need to add the cloud vendor list we've tested?

c3y1huang:
> LGTM. Do we need to add the cloud vendor list we've tested?

IMO no, because we are also not providing the setup for the cloud providers.

cc @innobead

derekbit (dismissed):

innobead:
> > LGTM. Do we need to add the cloud vendor list we've tested?
> 
> IMO no, because we are also not providing the setup for the cloud providers.
> 
> cc @innobead

We can just claim Longhorn is eligible for cluster autoscaler.

content/docs/1.3.0/high-availability/k8s-cluster-autoscaler.md
```diff
@@ -0,0 +1,22 @@
+---
+  title: Kubernetes Cluster Autoscaler Support (Experimental)
+  weight: 1
+---
+
+By default, Longhorn blocks Kubernetes Cluster Autoscaler from scaling down because:
```
innobead, line 6:
```suggestion
By default, Longhorn blocks Kubernetes Cluster Autoscaler from scaling down nodes because:
```

innobead (dismissed):
LGTM

PR #516, opened 2022-06-13
Add s390x IBM Z support

innobead:

### DIFF
--- a/content/docs/1.3.0/deploy/upgrade/longhorn-manager.md
+++ b/content/docs/1.3.0/deploy/upgrade/longhorn-manager.md
@@ -3,11 +3,11 @@ title: Upgrading Longhorn Manager
 weight: 1
 ---
 
-### Upgrading from v1.1.x or v1.2.x
+### Upgrading from v1.2.x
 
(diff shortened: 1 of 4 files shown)


content/docs/1.3.0/deploy/upgrade/longhorn-manager.md
```diff
@@ -3,11 +3,11 @@ title: Upgrading Longhorn Manager
 weight: 1
 ---
 
-### Upgrading from v1.1.x or v1.2.x
+### Upgrading from v1.2.x
 
-We only support upgrading to v{{< current-version >}} from v1.1.x or v1.2.x. For other versions, please upgrade to v1.1.x first.
+We only support upgrading to v{{< current-version >}} from v1.2.x. For other versions, please upgrade to v1.2.x first.
```
derekbit, line 8: Do we actually block the upgrade from v1.1.x in longhorn-manager?

derekbit (approved):
LGTM

content/docs/1.3.0/deploy/upgrade/longhorn-manager.md
innobead: No, we don't have that, so this is more like a convention but not strict from the implementation perspective.

PR #517, opened 2022-06-14
monitoring: add proxy gRPC connection metric

c3y1huang:
https://github.com/longhorn/longhorn/issues/4038


content/docs/1.3.0/monitoring/metrics.md
```diff
@@ -41,6 +41,7 @@ weight: 3
...
 | longhorn_instance_manager_memory_usage_bytes | The memory usage of this longhorn instance manager | longhorn_instance_manager_memory_usage_bytes{instance_manager="instance-manager-e-2189ed13",instance_manager_type="engine",node="worker-2"} 2.4072192e+07 |
 | longhorn_instance_manager_memory_requests_bytes | Requested memory in Kubernetes of this longhorn instance manager | longhorn_instance_manager_memory_requests_bytes{instance_manager="instance-manager-e-2189ed13",instance_manager_type="engine",node="worker-2"} 0 |
+| longhorn_instance_manager_proxy_grpc_connection | The number of proxy gRPC connection of this longhorn instance manager | longhorn_instance_manager_proxy_grpc_connection{container="longhorn-manager", endpoint="manager", instance="10.42.0.236:9500", instance_manager="instance-manager-e-814dfd05", instance_manager_type="engine", job="longhorn-backend", namespace="longhorn-system", node="worker-2", pod="longhorn-manager-tvqpc", service="longhorn-backend"} 0
```
innobead, line 44: I think we should just keep the labels we exported. For other general labels, I believe they are just about the pods which are monitored.

innobead (changes requested):

content/docs/1.3.0/monitoring/metrics.md
innobead: cc @PhanLe1010

innobead (approved):

PR #525, opened 2022-06-21
doc: Update the volume size doc

shuo-wu:
Longhorn/longhorn#4158


content/docs/1.3.0/volumes-and-nodes/volume-size.md
```diff
@@ -91,24 +105,37 @@ Here we summarize the important things related to disk space usage we have in th
...
         ```
-        (N + 1 + 1) x head/snapshot average actual size
+        (M + N + 1 + 1 + 1 + 1) x head/snapshot average actual size
         ```
 
-        where N is the number of snapshots retained, the 1st 1 is for the system snapshot, and the 2nd 1 is for the temporary space that may be required by snapshot deletion.
+        where `M` is the snapshots created by users manually, `N` is the snapshot recurring job retain number, the 1st `1` means the volume head, the 2nd `1` is the system snapshots created by rebuilding or expansion operations, the 3rd `1` is the extra snapshot before the recurring job starts cleanup, and the 4th `1` is for the temporary space that may be required by snapshot deletion.
+   - If users don't want snapshot at all. The formula would be
+       ```
+       (1 + 1 + 1) x head/snapshot average actual size
```
PhanLe1010, line 127: typo: missing one number 1

content/docs/1.3.0/volumes-and-nodes/volume-size.md
```diff
@@ -91,24 +105,37 @@ Here we summarize the important things related to disk space usage we have in th
...
+       ```
+
+       where the 1st `1` means the volume head, the 2nd `1` is the first system snapshot created by rebuilding or expansion operations, the 3rd `1` is the second system snapshot, and the 4th `1` is for the temporary space that may be required by snapshot deletion/merging the 1st system snapshot with the 2nd one.
```
PhanLe1010, line 130: This fomular is bigger than the number that we gave UiPath (3x). Should we mention that this is the absolute worst case? The arrange case should be 3x?

shuo-wu: Not sure. Typically 3x is large enough. This is actually the corner case:
1. Volume head is full
2. 1st Rebuilding is triggered. (The purge of this rebuilding cannot do anything.)
3. Before there is another pruning coming, the volume head suddenly becomes full and the 2nd rebuilding is triggered.
4. During the purge (for the 2nd rebuilding), the volume head becomes full

content/docs/1.3.0/volumes-and-nodes/volume-size.md
shuo-wu: Oh YES. Thank you for pointing that!

content/docs/1.3.0/volumes-and-nodes/volume-size.md
derekbit: One question: will the snapshot purging and rebuilding/expanding happen simultaneously?

shuo-wu: IIRC, No. The snapshot purge is the last step of the rebuilding, which happens after the new replica becomes RW.

innobead (commented):
LGTM. After discussing with @shuo-wu, just one part related to the no-user snapshot case, and we probably need to explain a bit why there will be 2 system snapshots following the volume head.

innobead (approved):
LGTM

innobead:
cc @longhorn/qa

PR #528, opened 2022-06-25
Update helm installation and upgrade doc

derekbit:
[Longhorn 4175](https://github.com/longhorn/longhorn/issues/4175)


innobead (approved):
LGTM

PR #542, opened 2022-07-22
Remove IOPS requirements of the volume configuration explanation 

innobead:


shuo-wu:
We may need to backport this to v1.1.4, v1.2.5, and v1.3.1.

innobead:
> We may need to backport this to v1.1.4, v1.2.5, and v1.3.1.

Yes, but 1.1.4 will not, because we only verified 1.24, and 1.3.0 and 1.1.4 will not be planned to release anymore.

PR #545, opened 2022-07-22
Update the doc for feature purging-before-rebuilding

shuo-wu:
https://github.com/longhorn/longhorn/issues/4183
https://github.com/longhorn/longhorn/issues/4251


innobead:
Just created v1.4.0 branch, we need to contribute there as well.

innobead (approved):
LGTM

innobead:
Just found this merge is incorrect. This should be for 1.3.1 instead of 1.3.0. 

@shuo-wu Please help fix this. Thanks.

PR #555, opened 2022-07-26
Update the v1.4.0 doc for feature purging-before-rebuilding

shuo-wu:
https://github.com/longhorn/longhorn/issues/4183


innobead (approved):

innobead:
nit: The explanation is good. Do you think this also deserves an illustration as well like others on the same page?

shuo-wu:
If there are users/customers feeling confused about this part, yes we can add illustrations.

PR #565, opened 2022-08-16
Fix wrong formatted json

hethehe:
delete unnecessary commas


innobead (approved):

innobead (changes requested):
Please also update other version docs as well. Thanks. 

Also, sign your commit to resolve the DCO check issue.

hethehe:
see #567

PR #574, opened 2022-09-02
Fix the flag for chart migration script

PhanLe1010:
longhorn/longhorn#3714

There shouldn't have the = character for the flags


PhanLe1010:
ping @innobead

innobead (approved):

PR #575, opened 2022-09-02
Release 1.2.5

innobead:


mantissahz (approved):

mantissahz:
Docsearch issue would be solved after tuning the config of Alogila and change the docsearch API version from v2 to v3.

innobead:
> Docsearch issue would be solved after tuning the config of Alogila and change the docsearch API version from v2 to v3.

Awesome! When will it be ready?

mantissahz:
After some tests, maybe this night or next Monday.

PR #576, opened 2022-09-05
Migrate docsearch from v2 to v3.

mantissahz:
Change css url and appId, apiKey.
Re-indexing the website at Algolia/Crawler website.

longhorn/longhorn#4457

Signed-off-by: James Lu <james.lu@suse.com>

### DIFF
--- a/layouts/partials/search-bar.html
+++ b/layouts/partials/search-bar.html
@@ -1,8 +1,5 @@
 <div class="navbar-item is-hidden-touch">
   <p class="control has-icons-left">
-    <input class="input is-rounded" type="text" placeholder="Search" id="search-input">
-    <span class="icon is-small is-left">
-      <ion-icon name="search-outline"></ion-icon>
-    </span>
+    <div id="docsearch"></div>
   </p>
 </div>
--- a/layouts/partials/css.html
+++ b/layouts/partials/css.html
@@ -12,4 +12,4 @@
 {{ $prodCss      := $css | fingerprint }}
 <link rel="stylesheet" href="{{ $prodCss.RelPermalink }}" integrity="{{ $prodCss.Data.Integrity }}">
 {{ end }}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3" />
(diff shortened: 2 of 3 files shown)


innobead:
@mantissahz After this, all work for non indexed versions?

mantissahz:
>After this, all work for non indexed versions?

@innobead I still try to automatically add `version` to the index result but it would not affect the frontend codes. It would be changed at the backend of the Algolia crawler.
Now I manually add the rules for each version now such as pages under `https://longhorn.io/docs/1.3.1/` would be indexed with `version: 1.3.1` and it would be filtered by frontend with `searchParameters: { 'facetFilters': ["version:{{ $currentVersion }}"] }`

layouts/partials/search-bar.html
```diff
@@ -1,8 +1,7 @@
 <div class="navbar-item is-hidden-touch">
   <p class="control has-icons-left">
-    <input class="input is-rounded" type="text" placeholder="Search" id="search-input">
-    <span class="icon is-small is-left">
-      <ion-icon name="search-outline"></ion-icon>
-    </span>
+    <div class="container">
+      <div id="docsearch"></div>
```
innobead, line 4: I guess this change is from the their example? But will this impact CSS styling change?

innobead:
> >After this, all work for non indexed versions?
> 
> 
> 
> @innobead I still try to automatically add `version` to the index result but it would not affect the frontend codes. It would be changed at the backend of the Algolia crawler.
> 
> Now I manually add the rules for each version now such as pages under `https://longhorn.io/docs/1.3.1/` would be indexed with `version: 1.3.1` and it would be filtered by frontend with `searchParameters: { 'facetFilters': ["version:{{ $currentVersion }}"] }`

IIRC, this is what other versions work after Jenting change, correct?

Anything new changed? Or anything I probably missed?

layouts/partials/search-bar.html
mantissahz: Yeah, It's from Docserach official website.
I am not good at web UI and need someone's double check, but it seams the deploy-preview looks the same as original one except the style of search bar.

innobead: You can run `make serve` to see locally, but I believe the style will be impacted. Or, we can check it together tomorrow.

mantissahz:
Just changed the backend config, and I think it is ready for all and non indexed versions.

layouts/partials/search-bar.html
innobead: After discussing, the style will be controlled by docsearch styling instead, so we don't need to worry about it.

innobead (approved):

layouts/partials/search-bar.html
mantissahz: Yeah, I did use `make serve` to check the changes.
Modified. Thanks. only `<div id="docsearch"></div>`

PR #583, opened 2022-09-30
Add analysis-filesystem-corrupted-issues-due-to-error-on-cow-while-rebuilding-replicas

PhanLe1010:
longhorn/longhorn#4354


content/kb/analysis-filesystem-corrupted-issues-due-to-error-on-cow-while-rebuilding-replicas.md
```diff
@@ -0,0 +1,196 @@
+---
+title: "Analysis: Potential Data/Filesystem Corruption"
+author: Derek Su and Phan Le
+draft: false
+date: 2022-09-29
+categories:
+- "data-integrity"
+---
+
```
joshimoo, line 9: This is good from a technical details points, let's add a small summary above the affected versions.

Makings sure that the user knows that this isn't very likely to happen to them (needs to be rebuilding, non 4k (default) block size) and if so they would have catched it most likely after the next mount.

PhanLe1010:
Putting this blog PR on hold until after we release the fixes (v1.3.2 and v1.2.6)

innobead (approved):
1.3.2 and 1.2.6 have been released. Let's publish this.

PR #588, opened 2022-10-03
Added text to advise against specifying resources in Helm charts

weizhe0422:
[Longhorn 4601](https://github.com/longhorn/longhorn/issues/4601)


innobead (commented):
This change will be possibly confusing for users, especially updated in this section. 

As discussed with other members, let's delay this explanation and see if we really need that until there will be further feedback from the community. 

WDYT?

weizhe0422:
I am agree with that.  Therefore I think this PR should be close/pending and make the [Issue#4601](https://github.com/longhorn/longhorn/issues/4601) keeps going.

PR #592, opened 2022-10-07
docs(deploy): Update the Environment Check Script output result

weizhe0422:
[Longhorn 4450](https://github.com/longhorn/longhorn/issues/4450)


innobead:
@weizhe0422 Please only update to the releases with this change, because IIRC not all versions  have this output.

weizhe0422:
Check the all versions of script, only the > 1.2.5 will show the new format output. 

Therefore only > 1.2.5 documents will be updated.

derekbit (approved):

innobead (approved):

PR #628, opened 2022-12-05
doc: add "Fast Rebuild Rebuild" doc and settings

derekbit:
Longhorn/longhorn#4783
    
### DIFF
--- a/content/docs/1.4.0/advanced-resources/fast-replica-rebuild.md
+++ b/content/docs/1.4.0/advanced-resources/fast-replica-rebuild.md
@@ -7,11 +7,13 @@ Longhorn supports fast replica rebuilding based on the checksums of snapshot dis
 
 ## Introduction
 
-The legacy replica rebuilding process walks through all snapshot disk files. For each data block, the client (healthy replica) hashes the local data block as well as requests the checksum of the corresponding data block in the remote side (rebuilt replica). Then, the client compares the two checksums to determine if the data block needs to be sent to the remote side and override the data block. Thus, it is an IO- and computing-intensive process, especially if the volume is large or contains a large number of snapshot files.
-If users enable the snapshot data integrity check feature by configuring `snapshot-data-integrity` to `enabled` or `fast-check`, the change time and the checksum of snapshot disk files are recorded. As long as the two conditions are met
-- The change timestamp on the snapshot disk file and the value recorded are the same
-- Both the local and remote snapshot disk files have the same checksum
-We can skip the synchronization of the snapshot disk file. Then, a reduction in the number of unnecessary computations can speed up the entire process as well as reduce the impact of the system performance.
+The legacy replica rebuilding process walks through all snapshot disk files. For each data block, the client (healthy replica) hashes the local data block as well as requests the checksum of the corresponding data block on the remote side (rebuilt replica). Then, the client compares the two checksums to determine if the data block needs to be sent to the remote side and override the data block. Thus, it is an IO- and computing-intensive process, especially if the volume is large or contains a large number of snapshot files.
+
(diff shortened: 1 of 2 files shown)


content/docs/1.4.0/advanced-resources/fast-replica-rebuild.md
```diff
@@ -0,0 +1,21 @@
+---
+title: Fast Replica Rebuild
+weight: 4
+---
+
+Longhorn supports fast replica rebuilding based on the checksums of snapshot disk files.
+
+## Introduction
+
+The legacy replica rebuilding process walks through all snapshot disk files. Data blocks are hashed in both the local and remote disks to determine if they need to be sent to the remote side. Thus, it is an IO- and computing-intensive process, especially if the volume is large or contains a large number of snapshot files.
```
innobead, line 10: IIRC, the data blocks of each snapshot will be always transferred from a remote replica, then they will be compared with the data blocks of the local snapshot to decide if they should be saved/overridden locally. If yes, then we need to explain this correctly.

derekbit: > IIRC, the data blocks of each snapshot will be always transferred from a remote replica

Client sends a request to server for requesting the remote block's checksum. The checksums are calculated in the remote and local sides, respectively.

innobead: Clarified, all good!

derekbit: I rephrased the sentences and make them more clear.

content/docs/1.4.0/advanced-resources/fast-replica-rebuild.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Introduction
+
+The legacy replica rebuilding process walks through all snapshot disk files. For each data block, the client (healthy replica) hashes the local data block as well as requests the checksum of the corresponding data block in the remote side (rebuilt replica). Then, the client compares the two checksums to determine if the data block needs to be sent to the remote side and override the data block. Thus, it is an IO- and computing-intensive process, especially if the volume is large or contains a large number of snapshot files.
+If users enable the snapshot data integrity check feature by configuring `snapshot-data-integrity` to `enabled` or `fast-check`, the change time and the checksum of snapshot disk files are recorded. As long as the two conditions are met
+- The change timestamp on the snapshot disk file and the value recorded are the same
+- Both the local and remote snapshot disk files have the same checksum
+We can skip the synchronization of the snapshot disk file. Then, a reduction in the number of unnecessary computations can speed up the entire process as well as reduce the impact of the system performance.
+
```
innobead, line 15:
```suggestion
The legacy replica rebuilding process walks through all snapshot disk files. For each data block, the client (healthy replica) hashes the local data block as well as requests the checksum of the corresponding data block on the remote side (rebuilt replica). Then, the client compares the two checksums to determine if the data block needs to be sent to the remote side and override the data block. Thus, it is an IO- and computing-intensive process, especially if the volume is large or contains a large number of snapshot files.

If users enable the snapshot data integrity check feature by configuring `snapshot-data-integrity` to `enabled` or `fast-check`, the change time and the checksum of snapshot disk files are recorded. As long as the two below conditions are met, we can skip the synchronization of the snapshot disk file. 
- The change timestamps on the snapshot disk file and the value recorded are the same
- Both the local and remote snapshot disk files have the same checksum

Then, a reduction in the number of unnecessary computations can speed up the entire replica rebuilding process as well as reduce the impact on the system performance.

```

innobead (changes requested):

innobead (approved):

PR #632, opened 2022-12-07
Update RWX volume doc

derekbit:
Longhorn/longhorn#2293


content/docs/1.4.0/advanced-resources/rwx-workloads.md
```diff
@@ -1,60 +1,69 @@
...
+{{< figure src="/img/diagrams/rwx/rwx-arch.png" >}}
 
 # Requirements
 
-To be able to use RWX volumes, each client node needs to have a NFSv4 client installed.
+It is necessary to meet the following requirements in order to use RWX volumes.
 
-For Ubuntu you can install a NFSv4 client via:
+1. Each NFS client node needs to have a NFSv4 client installed.
+     - For Debian based distros you can install a NFSv4 client via:
```
innobead, line 20: Can add a SUSE/openSUSE example. Also, there is a help manifest users can leverage.

content/docs/1.4.0/advanced-resources/rwx-workloads.md
```diff
@@ -1,60 +1,69 @@
...
-```
-yum install nfs-utils
-```
+2. The hostname of each node is unique in the Kubernetes cluster.
 
-If the NFSv4 client is not available on the node, when trying to mount the volume the below message will be part of the error:
-```
-for several filesystems (e.g. nfs, cifs) you might need a /sbin/mount.<type> helper program.\n
-```
+    There is a dedicated recovery backend service for NFS servers in Longhorn system. When a client connects to an NFS server, the client's information, including its hostname, will be stored in the recovery backend. When a share-manager Pod or NFS server is abnormally terminated, Longhorn will create a new one. Within the 90-seconds grace period, clients will reclaim locks using the client information stored in the recovery backend.
```
innobead, line 37:
```suggestion
    There is a dedicated recovery backend service for the NFS server in each share manager pod in the Longhorn system. When a client connects to an NFS server, the client's information, including its hostname, will be stored in the recovery backend. When a share-manager Pod or the internal NFS server is abnormally terminated, Longhorn will create a new one. Within the 90-seconds grace period, clients will reclaim locks using the client information stored in the recovery backend.
```

content/docs/1.4.0/advanced-resources/rwx-workloads.md
```diff
@@ -1,60 +1,69 @@
...
 
-Any failure of the share-manager Pod (volume failure, node failure, etc) will lead to the Pod being recreated and the volume's `remountRequestedAt` flag to be set, which will lead to the workload Pods being deleted and Kubernetes
-recreating them. This functionality depends on the setting of [Automatically Delete Workload Pod when The Volume Is Detached Unexpectedly,](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly)
-which by default is `true`. If the setting is disabled, the workload Pods might end up with `io errors` on RWX volume failures.
+2. If the DNS service goes down, share-manager Pod will not be able to communicate with longhorn-nfs-recovery-backend
```
innobead, line 64:
```suggestion
2. If the Kubernetes DNS service goes down, share-manager Pod will not be able to communicate with longhorn-nfs-recovery-backend
```

innobead (changes requested):

innobead (commented):
How about adding the recovery backend service in the arch diagram?

content/docs/1.4.0/advanced-resources/rwx-workloads.md
innobead: The hostname reported by the NFS server should be the hostname of the share manager pod which is the name of the share manager CR, so why does the hostname of each Node matter?

I assume the hostname will be the same after the share manager pod failed over because the name should be still the same as the share manager CR. 

Anything missing?

innobead: Yeah, you are right.

About `So the client id, that the NFS server received from the client, is actually the hostname of the client’s host`, if more than one pods mount the same RWX volume, what the client ID will be for each pod which is on the same node? How to differentiate them? (I assume the client ID is composed of the host address + <id>?)

derekbit: IIRC, because we use NFS4.1, so there is only one `session` between the server node and the client node. All connections between them use the same session. Before creating the session, the client id is exchanged. Thus, all clients on a node use the same client id.

innobead: I see, then it's clarified :+1:

derekbit: > I assume the client ID is composed of the host address + ?

Yes, a generic NFS client id is `host name + ip of the client host`. But we will encounter an [issue](https://github.com/longhorn/longhorn/issues/2293#issuecomment-1194072597) if the ` ip of the client host` is added, so the `ip of the client host` is removed and only uses the host name.

content/docs/1.4.0/advanced-resources/rwx-workloads.md
derekbit: Updated. Thank you.

content/docs/1.4.0/advanced-resources/rwx-workloads.md
derekbit: Updated. Thank you.

innobead (approved):

PR #657, opened 2023-01-04
fix typo

npu21:
```
 and it reflects the live data of the volume at the time at the time the backup was created.
```


innobead:
@npu21 Please sign off your commit next time, because all contributions need to pass the [DCO check](https://github.com/apps/dco). Thanks.

PR #669, opened 2023-03-01
Add pod-selector to avoid draining Longhorn webhook

PhanLe1010:
longhorn/longhorn#3304

### DIFF
--- a/content/docs/1.3.0/volumes-and-nodes/maintenance.md
+++ b/content/docs/1.3.0/volumes-and-nodes/maintenance.md
@@ -16,9 +16,9 @@ This section describes how to handle planned maintenance of nodes.
 
 1. Drain the node to move the workload to somewhere else.
 
-    You will need to use `--ignore-daemonsets` and `--pod-selector='app!=csi-attacher,app!=csi-provisioner'` options to drain the node.
+    You will need to use `--ignore-daemonsets` and `--pod-selector='app!=csi-attacher,app!=csi-provisioner,app!=longhorn-admission-webhook,app!=longhorn-conversion-webhook,app!=longhorn-driver-deployer'` options to drain the node.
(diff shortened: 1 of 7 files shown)


innobead (commented):
Do we still need driver deployer? 

BTW, currently the deployments of webhooks and CSI sidecars have replicas with pod anti-affinity, so if users do drain nodes one by one, probably it's not required to skip these pods?

PhanLe1010:
> Do we still need driver deployer?

Good catch. Thank you. Fixed

> BTW, currently the deployments of webhooks and CSI sidecars have replicas with pod anti-affinity, so if users do drain nodes one by one, probably it's not required to skip these pods?

We still have problem in single node cluster

innobead:
If a single node, why need to drain?

PhanLe1010:
> If a single node, why need to drain?

When the user upgrades Kubernetes. Drain to stop the workload first so that we don't crash the instance manager when there is active workload. What do you think?

PhanLe1010:
I am also planning to fix the driver deployer so that it doesn't restart the csi deployment when Kubernetes version change for the next version. What do you think?

innobead:
> I am also planning to fix the driver deployer so that it doesn't restart the csi deployment when Kubernetes version change for the next version. What do you think?

As discussed, we all agree with this change to avoid any unexpected CSI sidecars restart caused by the deployer restart during the upgrade.

In different use cases, the usage of draining can be clarified as follows:

- In a multiple-node cluster, if users drain nodes one by one, specific pod exclusion should not be required because CSI sidecars and webhooks already have replicas and pod anti-affinity placement.
- In a single-node cluster, the drain operation seems not quite meaningful because if users want to upgrade K8s, they should shut down all workarounds instead.

What we cannot control is how users perform the drain operation in parallel, so this is why pod exclusion is still needed in the current practice in the doc until a new mechanism introduced in https://github.com/longhorn/longhorn/issues/3304

innobead (approved):

PR #670, opened 2023-03-03
docs(system-restore): ignore backup target settings

c3y1huang:
- https://github.com/longhorn/longhorn/issues/5442
- https://github.com/longhorn/longhorn/issues/5473


innobead (changes requested):
Can we also mention the default config map for default settings will not be restored as well because we will only restore eligible setting resources?

c3y1huang:
> Can we also mention the default config map for default settings will not be restored as well because we will only restore eligible setting resources?

Updated.

innobead (approved):
LGTM.

cc @longhorn/dev @longhorn/qa

PR #674, opened 2023-03-13
Update LH Doc to correct kubectl command in uninstallation job example

roger-ryao:
issue https://github.com/longhorn/longhorn/issues/5540

Signed-off-by: Roger Yao <roger.yao@suse.com>

### DIFF
--- a/content/docs/1.4.0/deploy/uninstall/_index.md
+++ b/content/docs/1.4.0/deploy/uninstall/_index.md
@@ -51,7 +51,7 @@ helm uninstall longhorn -n longhorn-system
     clusterrolebinding.rbac.authorization.k8s.io/longhorn-uninstall-bind created
     job.batch/longhorn-uninstall created
 
-    $ kubectl get job/longhorn-uninstall -n default -w
+    $ kubectl get job/longhorn-uninstall -n longhorn-system -w
     NAME                 COMPLETIONS   DURATION   AGE
     longhorn-uninstall   0/1           3s         3s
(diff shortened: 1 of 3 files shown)


innobead (changes requested):
Please also help update the 1.3 doc as well.

roger-ryao:
> Please also help update the 1.3 doc as well.

Hi @innobead 
     This change, which occurred in the Longhorn repository at https://github.com/longhorn/longhorn/commit/89270bf0fabe40dcffe4eb8b345ca21326ae6b0c, was made after the release of version 1.4.0 and has not been backported to version 1.3.x.

innobead:
@roger-ryao please refer to https://github.com/longhorn/longhorn-manager/pull/1518#discussion_r994626494 about the context. Let's update the 1.3.3 (not released) and 1.3.2 docs as well. Just need to update the latest patch.

innobead (changes requested):
As per the last comment, need to update 1.3 docs.

innobead (approved):

PhanLe1010:
I think we don't need to update the doc for 1.3.3 because the uninstallation job is still in the default namespace in 1.3.3

The reason that we moved the uninstallation job into `longhorn-system` namespace was that PSP is not available in K8S 1.25. However, Longhorn 1.3.3 doesn't support K8S 1.25 so we don't need to move the uninstallation job into `longhorn-system` namespace which means we don't need to update the doc

What do you think? @innobead

innobead:
Sounds good 👌

PR #677, opened 2023-03-20
Update Node Maintenance and Kubernetes Upgrade Guide

PhanLe1010:
longhorn/longhorn#3304
longhorn/longhorn#5549
longhorn/longhorn#5585


content/docs/1.3.3/volumes-and-nodes/maintenance.md
```diff
@@ -71,14 +64,37 @@ To remove a node:
...
+
+With the assumption that node and disks are not being deleted/removed, the recommended upgrading guide is:
+1. You should cordon and drain a node before upgrading Kubernetes components on a node.
+   Draining instruction is similar to the drain instruction at [Updating the Node OS or Container Runtime](#updating-the-node-os-or-container-runtime)
+2. The drain `--timeout` should be big enough so that replica rebuildings on healthy node can finish.
+   The more Longhorn replicas you have on the draining node, the more time it takes for the Longhorn replicas to be rebuilt on other healthy nodes.
+   We recommending you to test and select a big enough value or set it to 0 (aka never timeout).
+3. The number of nodes doing upgrade at a time should be smaller than the number of Longhorn replicas for each volume.
+   This is so that a running Longhorn volume has at least one healthy replica running at a time.
+4. Set the setting [Node Drain Policy](../../references/settings#node-drain-policy) to `allow-if-replica-is-stopped` so that the drain is not blocked by the last healthy replica of a detached volume.
```
shuo-wu, line 94: We can suggest users have an appropriate value for setting `replica-replenishment-wait-interval`, which can cooperate with this policy and avoid some unnecessary replica rebuilding.

shuo-wu: FYI: Here is a request ticket that modify the default value of this setting to 1800:
https://github.com/harvester/harvester/issues/3671

PhanLe1010: Sorry, What do you think would be the appropriate value for `replica-replenishment-wait-interval` in this draining case? 

Usually, node will be uncordoned immediately after finishing draining so it shouldn't be more than 10 minutes

shuo-wu: I don't have such a number in my mind since I am not experienced in the upgrade part.
But we can ask users to check and modify it based on their env/estimation here.

PhanLe1010: From my testing, the node will be uncordoned immediately after the replica instance manager is successfully drained. So I think the default 10 mins should be more than enough. Do you think we should still have the note?

content/docs/1.3.3/deploy/important-notes/index.md
```diff
@@ -7,7 +7,9 @@ This page lists important notes for Longhorn v{{< current-version >}}.
 Please see [here](https://github.com/longhorn/longhorn/releases/tag/v{{< current-version >}}) for the full release note.
 
 ## Notes
-1. Please ensure your Kubernetes cluster is at least v1.18 and at most v1.24 before upgrading to Longhorn v{{< current-version >}} because the supported Kubernetes version has been updated (>= v1.18 and <= v1.24) in v{{< current-version >}}.
+1. We introduce a new setting [Node Drain Policy](../../references/settings#node-drain-policy) which has some overlap with the old setting [Allow Node Drain with the Last Healthy Replica](../../references/settings#allow-node-drain-with-the-last-healthy-replica), so it might lead to some confusion.
```
innobead, line 10:
```suggestion
1. We introduce a new setting [Node Drain Policy](../../references/settings#node-drain-policy) which will deprecate the old setting [Allow Node Drain with the Last Healthy Replica](../../references/settings#allow-node-drain-with-the-last-healthy-replica), so it is recommended to keep the old setting [Allow Node Drain with the Last Healthy Replica](../../references/settings#allow-node-drain-with-the-last-healthy-replica) as default `false` value and use the new setting [Node Drain Policy](../../references/settings#node-drain-policy) which has more options and controls.
```

content/docs/1.3.3/deploy/important-notes/index.md
```diff
@@ -7,7 +7,9 @@ This page lists important notes for Longhorn v{{< current-version >}}.
...
+   To avoid the confusion, it is recommended to keep the old setting [Allow Node Drain with the Last Healthy Replica](../../references/settings#allow-node-drain-with-the-last-healthy-replica) as default `false` value and use the new setting [Node Drain Policy](../../references/settings#node-drain-policy) which has more option and controls.
```
innobead, line 11:
```suggestion
```

content/docs/1.3.3/references/settings.md
```diff
@@ -72,6 +73,17 @@ By default, Longhorn will block `kubectl drain` action on a node if the node con
 
 If this setting is enabled, Longhorn will not block `kubectl drain` action on a node even if the node contains the last healthy replica of a volume.
 
+#### Node Drain Policy
+
+> Default: `block-if-contains-last-replica`
```
innobead, line 78: Please mark `Allow Node Drain with the Last Healthy Replica` is deprecated.

content/docs/1.3.3/volumes-and-nodes/maintenance.md
```diff
@@ -1,59 +1,52 @@
...
 
 1. Drain the node to move the workload to somewhere else.
 
-    You will need to use `--ignore-daemonsets` and `--pod-selector='app!=csi-attacher,app!=csi-provisioner,app!=longhorn-admission-webhook,app!=longhorn-conversion-webhook,app!=longhorn-driver-deployer'` options to drain the node.
-    The `--ignore-daemonsets` is needed because Longhorn deployed some daemonsets such as `Longhorn manager`, `Longhorn CSI plugin`, `engine image`.
-    The `--pod-selector='app!=csi-attacher,app!=csi-provisioner,app!=longhorn-admission-webhook,app!=longhorn-conversion-webhook,app!=longhorn-driver-deployer'` is needed so that Longhorn can properly detaches Longhorn volumes (see the [GitHub issue](https://github.com/longhorn/longhorn/issues/3304) for more detail).
+   You will need to use `--ignore-daemonsets` and `--force` to drain the node.
+   The `--ignore-daemonsets` is needed because Longhorn deployed some daemonsets such as `Longhorn manager`, `Longhorn CSI plugin`, `engine image`.
+   The `--force` is needed because some Longhorn pods (instance manager and share manager pods) are not managed by a ReplicationController, Job, or DaemonSet.
```
innobead, line 21: Do you mean right now without `--force` the drain will fail?

innobead (changes requested):
In general, LGTM, but some wording needs to adjust as I mentioned in the comment. Please make the changes to all version docs.

content/docs/1.3.3/volumes-and-nodes/maintenance.md
PhanLe1010: Yes, without --force the drain will fail

content/docs/1.3.3/deploy/important-notes/index.md
PhanLe1010: Thank you. Fixed

content/docs/1.3.3/deploy/important-notes/index.md
PhanLe1010: Thank you. Fixed

content/docs/1.3.3/references/settings.md
PhanLe1010: I thought we agreed that we only deprecated it in Longhorn 1.5.0? What do you think?

innobead: yeah, as discussed, there is one thing that needs clarification. 

We can deprecate a setting or function in a patch version. For example, in this case, we will deprecate it in versions 1.3.3 and 1.4.2, and it should be removed from version 1.5.0.

https://github.com/longhorn/longhorn/wiki/Deprecation-Policy

```
- Deprecate a feature in a patch or feature release. The feature should still work until it is removed later.
- After deprecation, remove the feature in the just following feature release with the upgrade migration to ensure a smooth upgrade.
```

cc @longhorn/dev

content/docs/1.3.3/volumes-and-nodes/maintenance.md
innobead: cc @longhorn/qa

content/docs/1.3.3/volumes-and-nodes/maintenance.md
innobead: If the default value is good enough, then we don't need to emphasize another value here.

content/docs/1.3.3/references/settings.md
PhanLe1010: Thank you very much. All clear now. I marked it as deprecated in v1.3.3 and v1.4.2 and will remove it in v1.5.0 by this ticket https://github.com/longhorn/longhorn/issues/5620

innobead: Exactly @PhanLe1010 :+1:

innobead (approved):

content/docs/1.3.3/volumes-and-nodes/maintenance.md
khushboo-rancher: We have tested v1.3.3 drain without `force`, it seemed to have worked. https://longhorn.github.io/longhorn-tests/manual/pre-release/node/node-drain-deletion/.  We have also tested drain using Rancher UI, it seems without selecting `force` option, we were able to drain successfully.

innobead: @PhanLe1010 had a good comment for the context.

[/datastore/longhorn.go#L2775-L2777](https://github.com/longhorn/longhorn-manager/blob/b1b4fc0515ac34c43601c3c6503ac603e6ba9b54/datastore/longhorn.go#L2775-L2777)
```go
			// This field is needed so that `kubectl drain` can work without --force flag
			// See https://github.com/longhorn/longhorn/issues/1286#issuecomment-623283028 for more details
			Controller:         &controller,
```

PR #680, opened 2023-03-23
docs: Add KB about failure to delete orphaned pod volume directory

weizhe0422:
[Longhorn 3207](https://github.com/longhorn/longhorn/issues/3207)

### DIFF
--- a/content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
+++ b/content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
@@ -1,0 +1,42 @@
+---
+title: "Troubleshooting: Failure to delete orphaned Pod volume directory"
+author: Ray Chang
+draft: false
+date: 2023-07-20
+categories:
+  - "csi"
+---
+
+## Applicable versions
+
+All Longhorn versions.
+
+Kubernetes versions before `v1.28`. A backported PR to `v1.27` is awaiting merging.
+
+## Symptoms
+
+In the event of a worker node failure, while hosting active Pods, the Pods are gracefully evicted as the node undergoes downtime and awaits restoration. During this period, the kubelet, which is responsible for managing the node, will generate the following error messages at regular intervals of two seconds.
+```
+orphaned pod <pod-uid> found, but error not a directory occurred when trying to remove the volumes dir
+```
+
+## Reason
+
+This situation occurs when a node goes through a downtime state and then takes some time before entering the recovery phase. During this process, the affected Pods are evicted and relocated to other nodes. However, due to the disruption, the connection between the kubelet and the longhorn-csi-plugin is severed. Therefore, the kubelet encountered difficulties when deleting the `vol_data.json` file. This process is used to perform self-housekeeping tasks to clean up orphan volume mount points associated with evicted Pods. The kubelet, while capable of removing directories, cannot delete individual files, resulting in incomplete cleanup in this specific situation. [(source code)](https://github.com/kubernetes/kubernetes/blob/8c1dc65da905d0c8435659424169846ba2fb2d63/pkg/kubelet/kubelet_volumes.go#L155-L162).
+
+## Solution
+
+Once the node and kubelet have been restored, the longhorn-csi-plugin will automatically restart, allowing the Pod to remount the volume and resume its running state.
+
+However, in cases where the Pod and its associated volume are rescheduled to a different node, leaving behind a lingering `vol_data.json` file on the crashed node, manual intervention is required. You will need to manually delete the `vol_data.json` file located within the `/var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/pvc_<pod-uid>/` directory.
+
+Within the present Kubernetes master branch, the issue is addressed in version `1.28.x`, thereby ensuring that orphaned Pod volume mount points are properly cleaned up within the reconciliation loop. Moreover, a PR addressing the issue has been backported to version `1.27` and is presently awaiting the merging process.
+
+## Related information
+
+* Related Longhorn issue: https://github.com/longhorn/longhorn/issues/3207
+* Related Kubernetes issues & PRs
+   - https://github.com/kubernetes/kubernetes/issues/105536
+   - https://github.com/kubernetes/kubernetes/issues/111933
+   - master `v1.28.x`: https://github.com/kubernetes/kubernetes/pull/116134
+   - backport `v1.27`: https://github.com/kubernetes/kubernetes/pull/117235


content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,35 @@
...
+
+When a worker node crashes with some pods running on it, many error messages as shown below keep appearing in the kubelet.
+
+```
+orphaned pod <pod-uid> found, but error not a directory occurred when trying to remove the volumes dir
+```
+
+## Reason
+
+This is caused by the kubelet cleaning up volume folders but disconnecting from the CSI driver when the node crashes. Therefore, when the kubelet tries to delete `vol_data.json` to unmount the volume, it fails and causes further deletion of the old volume mount point to fail.
```
PhanLe1010, line 24: > when the kubelet tries to delete `vol_data.json` to unmount the volume

Do you mean that when the kubelet tries to teardown a volume by sending CSI request to longhorn-csi-plugin, it fails because longhorn-csi-plugin is not running. Then kubelet cannot delete `vol_data.json` to unmount the volume as the csi call failed.

weizhe0422: Yes, I think it's because the connection between kubelet and longhorn-csi-plugin is disconnected due to an abnormal node, but it tries to tear down the volume, which fails to delete `vol_data.json`.

ejweber:
@PhanLe1010 @weizhe0422 

This one is holding up https://github.com/longhorn/longhorn/issues/3207, which is being asked about internally. What do we need to do to get unstuck?

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,35 @@
...
+## Solution
+
+After the node and kubelet are restored, the longhorn-csi-plugin will automatically restart, and the pod with the new volume mounted will be at Running state again.
+
+However, if the replica is scheduled to another node and produces a dangling `vol_data.json` file on the crashed node already, then you need to manually delete `vol_data.json` which is under `/var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/pvc_<pod-uid>/` directory.
+
+## Related information
+
+* Related Longhorn issue: https://github.com/longhorn/longhorn/issues/3207
+* Related Kubernetes issue: https://github.com/kubernetes/kubernetes/issues/105536
```
ejweber, line 35: Kubernetes PR https://github.com/kubernetes/kubernetes/pull/116134 supposedly fixes the tracked issue for Kubernetes v1.27.0+, though the tracked issue hasn't been verified or closed. Can we mention that we only expect this issue to occur in Kubernetes <v.1.27 somewhere in this KB?

ejweber (changes requested):

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,35 @@
...
+
+When a worker node crashes with some pods running on it, many error messages as shown below keep appearing in the kubelet.
+
+```
+orphaned pod <pod-uid> found, but error not a directory occurred when trying to remove the volumes dir
+```
+
+## Reason
+
+This is caused by the kubelet trying to tear down the volume by sending a CSI request to the longhorn-csi-plugin, but a node failure resulted in a disconnect. Then, when the kubelet tries to delete `vol_data.json` to unmount the volume, it also failed due to the failure of the CSI request.
```
shuo-wu, line 24: IMO, this statement does not clarify the issue...
Isn't the root cause that the **invalid mount point** caused by the node crash would fail the **volume reconstruction** (upstream bug) during the node/kubelet recovery?

PhanLe1010: @shuo-wu Could you link me to the document describing this root cause? 

From the comment of @derekbit, it is mentioning that the root cause is the disconnection between kubelet and longhorn-csi-plugin https://github.com/longhorn/longhorn/issues/3207#issuecomment-974845277. So kubelet cannot tell longhorn-csi-plugin to clean up that stale `volume_data.json` file

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Solution
+
+After the node and kubelet are restored, the longhorn-csi-plugin will automatically restart, and the pod with the new volume mounted will be at Running state again.
+
+However, if the replica is scheduled to another node and produces a dangling `vol_data.json` file on the crashed node already, then you need to manually delete `vol_data.json` which is under `/var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/pvc_<pod-uid>/` directory.
```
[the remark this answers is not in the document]
PhanLe1010: Agree. How about a term that is more familiar with users: pod and its volume move to new node?

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
shuo-wu: Oh sorry, I didn't notice this explanation. My opinion on this issue is from the upstream fix: It removes the existence check of mount point for filesystem volume or symbolic link for block volume.
But here @derekbit confirmed that it's caused by the CSI plugin connection breaking... Not sure which one is the root

innobead: I feel both are the same thing, but the node hard reboot/crash is just one case of `the disconnection between kubelet and longhorn-csi-plugin`.

@weizhe0422 can base on the reviews here fine-tune the PR.

weizhe0422: I reference to https://github.com/kubernetes/kubernetes/issues/111933 and https://github.com/kubernetes/kubernetes/pull/116134 and think the situation is lead caused by the connection broken between kubelet and plug-in, and the kubelet will try to reconstruct when resuming but failed since the volume mount point on the reboot node is out of date. 
https://github.com/kubernetes/kubernetes/blob/b2f26460b9354b38b174a21cdbf15c391d63dade/pkg/kubelet/volumemanager/reconciler/reconstruct_common.go#L296-L303
```golang
isExist, checkErr := rc.operationExecutor.CheckVolumeExistenceOperation(volumeSpec, checkPath, volumeSpec.Name(), rc.mounter, uniqueVolumeName, volume.podName, pod.UID, attachablePlugin)
if checkErr != nil {
            return nil, checkErr
}
// If mount or symlink doesn't exist, volume reconstruction should be failed
if !isExist {
            return nil, fmt.Errorf("volume: %q is not mounted", uniqueVolumeName)
 }
```
Therefore, the mount point information cannot be properly updated so that it is cleaned up in the next reconcile cycle.
https://github.com/kubernetes/kubernetes/blob/16534deedf1e3f7301b20041fafe15ff7f178904/pkg/kubelet/volumemanager/reconciler/reconstruct_new.go#L155-L174
```golang
func (rc *reconciler) cleanOrphanVolumes() {
            if len(rc.volumesFailedReconstruction) == 0 {
                        return
            }


            for _, volume := range rc.volumesFailedReconstruction {
                        if rc.desiredStateOfWorld.VolumeExistsWithSpecName(volume.podName, volume.volumeSpecName) {
                                    // Some pod needs the volume, don't clean it up and hope that
                                    // reconcile() calls SetUp and reconstructs the volume in ASW.
                                    klog.V(4).InfoS("Volume exists in desired state, skip cleaning up mounts", "podName", volume.podName, "volumeSpecName", volume.volumeSpecName)
                                    continue
                        }
                        klog.InfoS("Cleaning up mounts for volume that could not be reconstructed", "podName", volume.podName, "volumeSpecName", volume.volumeSpecName)
                        rc.cleanupMounts(volume)
            }


            klog.V(2).InfoS("Orphan volume cleanup finished")
            // Clean the cache, cleanup is one shot operation.
            rc.volumesFailedReconstruction = make([]podVolume, 0)
}
```

weizhe0422: I'll update the PR to make it more clear based on the comments.

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
weizhe0422: Thanks, I also found out that this is asked to backported to 1.27 but havn't got merge yet (https://github.com/kubernetes/kubernetes/pull/117235). But I think I could update this PR to include relevant information.

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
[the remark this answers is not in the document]
weizhe0422: Thank you and I will update the description to make it correctly and clearly.

weizhe0422:
@ejweber @PhanLe1010 @shuo-wu @derekbit I've pushed an updated version, modifying the description and adding information about the upstream fix PR. I would like ask for help to review it again. Thank you.

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,39 @@
...
+author: Ray Chang
+draft: false
+date: 2023-03-23
+categories:
+  - "csi"
+---
+
+## Applicable versions
+
+All Longhorn versions.
```
ejweber, line 12: NIT: This heading is probably most often used to talk about Longhorn version (as you have done), but it might also be a good location to highlight Kubernetes version.
```suggestion
All Longhorn versions.
Kubernetes versions before v1.27.
```

weizhe0422: Thanks. I've updated to include the K8S version information in that block.

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,40 @@
...
+```
+orphaned pod <pod-uid> found, but error not a directory occurred when trying to remove the volumes dir
+```
+
+## Reason
+
+This situation occurs when a node goes through a downtime state and then takes some time before entering the recovery phase. During this process, the affected Pods are evicted and relocated to other nodes. However, due to the disruption, the connection between the kubelet and the longhorn-csi-plugin is severed. Therefore, the kubelet encountered difficulties when deleting the `vol_data.json` file. This process is used to perform self-housekeeping tasks to clean up orphan volume mount points associated with evicted Pods. The kubelet, while capable of removing directories, cannot delete individual files, resulting in incomplete cleanup in this specific situation. [(source code)](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_volumes.go#L155-L162).
```
shuo-wu, line 24: NIT: It's better to use a permalink instead. https://github.com/kubernetes/kubernetes/blob/8c1dc65da905d0c8435659424169846ba2fb2d63/pkg/kubelet/kubelet_volumes.go#L155-L162

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Solution
+
+Once the node and kubelet have been restored, the longhorn-csi-plugin will automatically restart, allowing the Pod to remount the volume and resume its running state.
+
+However, in cases where the Pod and its associated volume are rescheduled to a different node, leaving behind a lingering `vol_data.json` file on the crashed node, manual intervention is required. You will need to manually delete the `vol_data.json` file located within the `/var/lib/kubelet/pods/<pod-uid>/volumes/kubernetes.io~csi/pvc_<pod-uid>/` directory.
+
+For Kubernetes versions 1.27.0 and above, this issue is addressed, ensuring that orphaned Pod volume mount points are properly cleaned up within the reconciliation loop.
```
shuo-wu, line 32: The backporting fix PR for v1.27 is not merged yet. Can we say this now?

shuo-wu (approved):
In general LGTM

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
weizhe0422: Got it, I've updated it. thanks.

content/kb/troubleshooting-failure-to-delete-orphaned-pod-volume-directory.md
weizhe0422: Yes, the [backport PR for v1.27](https://github.com/kubernetes/kubernetes/pull/117235) has not been merged yet. I would like to consult whether it would be more suitable if I modify it like this:

Within the present Kubernetes master branch, the issue is addressed, thereby ensuring that orphaned Pod volume mount points are properly cleaned up within the reconciliation loop. Moreover, a PR addressing the issue has been backported to version `1.27` and is presently awaiting the merging process.

ejweber: A good point. Maybe this makes my request to update the Applicable versions section premature.

weizhe0422: Thanks. I updated this paragraph as well as the "Applicable Versions" section.

ejweber (approved):
LGTM

PR #681, opened 2023-03-23
Update doc 'networking.md'

shuo-wu:
Longhorn/longhorn#5651


content/docs/1.3.0/references/networking.md
```diff
@@ -28,8 +28,9 @@ From | Port | Protocol
 To | Port | Protocol
 --- | --- | ---
 `Other Longhorn Manager` | 9500 | TCP
-`Instance Manager` |  8500; 10000-30000 | TCP
-`Backing Image Manager` |  8000; 8001 | TCP
+`Instance Manager` | 8500; 8501; 10000-30000 | TCP
```
shuo-wu, line 31: Will Longhorn managers directly talk with engine/replica processes after introducing the proxy? @c3y1huang

PhanLe1010: Is the port `8501` for proxy grpc calls from longhorn-manager to the instance manager?

content/docs/1.3.0/references/networking.md
```diff
@@ -28,8 +28,9 @@ From | Port | Protocol
...
+`Backing Image Manager` | 8000; 8001 | TCP
+`Backing Image Data Source` | 8001; 8002 | TCP
```
PhanLe1010, line 33: Could you point me to the use pf 8002? Thank you

content/docs/1.5.0/references/networking.md
```diff
@@ -45,16 +46,16 @@ To | Port | Protocol
 #### ingress
 From | Port | Protocol
 --- | --- | ---
-`Longhorn Manager` | 8500; 10000-30000 | TCP
+`Longhorn Manager` | 8500; 8501 | TCP
 `Other Instance Manager` | 10000-30000 | TCP
 `Node in the Cluster` | 3260 | TCP
-`Backing Image Manager` | 10000-30000 | TCP
+`Backing Image Data Source` | 10000-30000 | TCP
```
PhanLe1010, line 52: I forgot why instance-manager needs to talk to the backing image manager. Is it for exporting volume?

content/docs/1.3.0/references/networking.md
shuo-wu: Yes. I missed this update.

content/docs/1.3.0/references/networking.md
shuo-wu: It's used by the sync server/receiver of volume export:
https://github.com/longhorn/backing-image-manager/blob/455f018e5a80943c5d93a5f216ea973740f9fcee/pkg/datasource/server.go#L22-L26

content/docs/1.5.0/references/networking.md
shuo-wu: I think so. I didn't find any other case. 
And now volume export is handled by Backing Image Data Source instead.

content/docs/1.3.0/references/networking.md
c3y1huang: > Will Longhorn managers directly talk with engine/replica processes after introducing the proxy? @c3y1huang

Longhorn managers still manage processes through the 8500 port.

shuo-wu: How about port 10000-30000? Can we remove this? (Although I already removed this from the instance manager ingress part.)

c3y1huang: I am not entirely sure. Proxy is not responsible for handling tgt processes and their connections. And as far as I know, the process connections are to the local iSCSId from the IM-e, so I am not sure why the Manager needs to have those ports on egress.

shuo-wu: longhorn managers will handle the instance managers via 8500. As for the tgt part, it's the responsibility of the engine processes inside instance managers. I think we can remove it first and see the test result.

content/docs/1.5.0/references/networking.md
```diff
@@ -28,8 +28,9 @@ From | Port | Protocol
 To | Port | Protocol
 --- | --- | ---
 `Other Longhorn Manager` | 9500 | TCP
-`Instance Manager` |  8500; 10000-30000 | TCP
-`Backing Image Manager` |  8000; 8001 | TCP
+`Instance Manager` | 8500; 8501 | TCP
+`Backing Image Manager` | 8000; 8001 | TCP
+`Backing Image Data Source` | 8001; 8002 | TCP
```
ChanYiLin, line 33: I think Backing Image Data Source server and sync servers' port are `8000` and  `8001`
ref:
https://github.com/longhorn/backing-image-manager/blob/32cc0a79d209d1f6c2bd8323061e82f80b60970b/app/cmd/data_source.go#L22-L29

content/docs/1.5.0/references/networking.md
```diff
@@ -45,16 +46,16 @@ To | Port | Protocol
 #### ingress
 From | Port | Protocol
 --- | --- | ---
-`Longhorn Manager` | 8500; 10000-30000 | TCP
+`Longhorn Manager` | 8500; 8501 | TCP
 `Other Instance Manager` | 10000-30000 | TCP
```
ChanYiLin, line 50: I found the syncAgent in longhorn-engine would listen on `9504`
and use `9700-9800` for syncing files between replicas
should also be added to `other Instance Manager`
https://github.com/longhorn/longhorn-engine/blob/master/app/cmd/sync_agent.go#L21-L33

content/docs/1.5.0/references/networking.md
shuo-wu: It seems that the Longhorn manager does not need to send requests to the sync server (port 8001) of Backing Image Data Source pods as well. The Longhorn managers always talk with data source servers then data source servers proxies the requests to the sync server.

Similar for the Backing Image Manager.

In brief, Longhorn manager egress should be:
......
`Backing Image Manager` | 8000 | TCP
`Backing Image Data Source` | 8001 | TCP

content/docs/1.5.0/references/networking.md
shuo-wu: > start a replica client to ask instance manger to export the volume

This means that the BIDS pod will send a file sync request to the replica process inside the instance manager pod:
https://github.com/longhorn/longhorn-manager/blob/b9ff9347ed259338340211a62730e6c3261e42bb/controller/backing_image_data_source_controller.go#L765
Here, the port of the replica address is range 10000-30000.

content/docs/1.5.0/references/networking.md
shuo-wu: Currently, this cmd is not used by anyone. The sync agent server will be launched along with replica servers, and replica servers use port 10000-30000

shuo-wu:
@ChanYiLin @PhanLe1010 Can you help re-check this?

ChanYiLin (approved):

ChanYiLin:
LGTM!

derekbit (approved):
LGTM

shuo-wu:
@innobead Can you check this PR? Then we can update the network policy example and include the change in v1.3.3.

PR #693, opened 2023-04-27
doc: add doc for csi volumesnapshot associated with longhorn backing image

ChanYiLin:
ref: [longhorn/longhorn#5005](https://github.com/longhorn/longhorn/issues/5005)

### DIFF
--- a/content/docs/1.5.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backing-image.md
+++ b/content/docs/1.5.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backing-image.md
@@ -1,0 +1,221 @@
+---
+title: CSI VolumeSnapshot Associated with Longhorn BackingImage
+weight: 2
+---
+
+BackingImage in Longhorn is an object that represents a QCOW2 or RAW image which can be set as the backing/base image of a Longhorn volume.
+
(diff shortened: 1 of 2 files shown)


content/docs/1.5.0/snapshots-and-backups/csi-snapshot-support/csi-volume-snapshot-associated-with-longhorn-backing-image.md
```diff
@@ -0,0 +1,225 @@
...
+    volumeSnapshotContentName: test-on-demand-backing
+```
+
+Now you can create a `PerstistantVolumeClaim` object that refers to the newly created `VolumeSnapshot` object.
+Longhorn will create the BackingImage with the parameters provide in the `snapshotHandle`.
+For an example see [Restore PVC from CSI VolumeSnapshot Associated With Longhorn BackingImage](#restore-pvc-from-csi-volumesnapshot-associated-with-longhorn-backingimage) above.
+
+
+## Current Limitation
+* Longhorn volume must be in the attached state to create/restore BackingImage using CSI VolumeSnapshot mechanism.
```
innobead, line 225: Don't we have auto attach/detach to handle CSI snapshot? Is this limitation still valid?

innobead (changes requested):
In general, LGTM. Just one question needed to answer.

innobead (approved):

PR #695, opened 2023-05-10
doc: auto pod deletion setting does not support rwx volume

innobead:


innobead:
cc @longhorn/qa @derekbit

PR #696, opened 2023-05-11
doc(setting): allow collecting longhorn usage metrics

c3y1huang:
https://github.com/longhorn/longhorn/issues/5235


content/docs/1.5.0/references/settings.md
```diff
@@ -225,7 +226,65 @@ Upgrade Checker will check for a new Longhorn version periodically. When there i
...
+- Number of Longhorn nodes.
+- Number of volumes of each access mode (RWO, RWX, unknown).
+- Number of volumes of each data locality type (disabled, best_effort, strict_local, unknown).
+- Number of volumes of each frontend type (blockdev, iscsi).
+- Average volume size.
+- Average volume actual size.
+- Average number of snapshots per volume.
+- Average number of replicas per volume.
+- Average Longhorn component CPU usage (engine image, instance manager, manager, ui)
+- Average Longhorn component CPU usage (engine image, instance manager, manager, ui)
```
innobead, line 257: I would suggest let's collect control-plane (longhorn-manager) and data-plane (instance manager) only. 

Why there are two CPU usage? I think one is the CPU request set, right?

innobead (changes requested):
In general, LGTM. Just one comment.

content/docs/1.5.0/references/settings.md
c3y1huang: > I would suggest let's collect control-plane (longhorn-manager) and data-plane (instance manager) only.

Updated.

> Why there are two CPU usage? I think one is the CPU request set, right?

Meant memory. Corrected!

content/docs/1.5.0/references/settings.md
```diff
@@ -213,7 +214,62 @@ Upgrade Checker will check for a new Longhorn version periodically. When there i
...
+    - Current Longhorn Version
+    - Latest Longhorn Version
+    - Stable Longhorn Versions
+    - Default Longhorn Static Storage Class
+    - Deleting Confirmation Flag
+    - Default Data Path
+    - Upgrade Checker
+    - Allow Collecting Longhorn Usage
+  - Partially included:
+    - Backup Target Type/Protocol (cifs, nfs, s3, none, unknown). This is from the Backup Target setting.
```
innobead, line 262: Also azure.

innobead (changes requested):
Just one missing for azure backup target.

innobead (approved):
LGTM.

PR #703, opened 2023-05-17
doc: remove default image global settings

mantissahz:
Remove the global default image settings from v1.5.0 including `default-instance-manager-image` and `default-share-manager-image`.

Ref: longhorn/longhorn#5028


innobead (approved):

innobead (changes requested):
@mantissahz Please also update the important note about this removal.

mantissahz:
Added the section for removing default manager image settings in the important note

innobead (approved):

PR #704, opened 2023-05-19
Update auto-upgrade-engine.md

Tejeev:
The volume doesn't upgrade the engine, but it is the volume's Engine that is upgraded (if I understood correctly).  Corrected that and changed some of the wording to be easier to follow. Please review to be sure I understood correctly

### DIFF
--- a/content/docs/1.4.2/deploy/upgrade/auto-upgrade-engine.md
+++ b/content/docs/1.4.2/deploy/upgrade/auto-upgrade-engine.md
@@ -38,8 +38,8 @@ So, Longhorn leaves it to you to decide when it is the good time to manually upg
 At this time, Longhorn will automatically do offline upgrade for the volume similar to the detached volume case.
 
 #### 3. What Happened If The Upgrade Fails?
-If a volume failed to upgrade its engine, the engine image in volume's spec will remain to be different than the engine image in the volume's status.
-Longhorn will continuously retry to upgrade until it succeeds.
+If a volume's engine fails to upgrade, the volume's status won't change. 
+Because the status will remain different from the spec, Longhorn will retry the upgrade until it succeeds.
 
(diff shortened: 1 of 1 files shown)


content/docs/1.4.2/deploy/upgrade/auto-upgrade-engine.md
```diff
@@ -38,8 +38,8 @@ So, Longhorn leaves it to you to decide when it is the good time to manually upg
 At this time, Longhorn will automatically do offline upgrade for the volume similar to the detached volume case.
 
 #### 3. What Happened If The Upgrade Fails?
-If a volume failed to upgrade its engine, the engine image in volume's spec will remain to be different than the engine image in the volume's status.
-Longhorn will continuously retry to upgrade until it succeeds.
+If a volume's engine fails to upgrade, the volume's status won't change. 
+Because the status will remain different from the spec, Longhorn will retry the upgrade until it succeeds.
```
innobead, line 42:
```suggestion
Because the status of the engine image will remain different from the spec, Longhorn will retry the upgrade until it succeeds.
```

innobead (changes requested):
I don't see any obvious issues from the original explanation, but your update is also OK to me. Just a minor change. 

Also, please also make the change to all supported versions including unreleased ones. Thanks.

innobead:
@Tejeev we have DCO check, so remember to sign off your commit via `git commit -s`

content/docs/1.4.2/deploy/upgrade/auto-upgrade-engine.md
Tejeev: The issue with the original is that it leaves much implicit rather than explicit.  Customers were confused.
How do I fix other versions?

Tejeev:
> @Tejeev we have DCO check, so remember to sign off your commit via `git commit -s`

What is a DCO check? I made the pr through github.

PR #705, opened 2023-05-19
doc(recurring-job): set to PVC

c3y1huang:
https://github.com/longhorn/longhorn/issues/5791


innobead (approved):
LGTM.

innobead:
@c3y1huang just needs to ensure upgrade is compatible with existing labeled volumes.

c3y1huang:
> @c3y1huang just needs to ensure upgrade is compatible with existing labeled volumes.

Test result ref: https://github.com/longhorn/longhorn-manager/pull/1911#issuecomment-1556358677

PR #706, opened 2023-05-19
Slightly tweak Kubernetes minimum version wording

ejweber:
longhorn/longhorn#5595

It looks like we will not need to change the minimum Kubernetes version for Longhorn v1.5. Slightly tweak the wording on the website (since the supported version has not changed).


PhanLe1010 (approved):
LGTM

innobead (approved):

PR #707, opened 2023-05-23
Update _index.md

Tejeev:
Our CRD cleanup doesn't mention the webhooks which should be cleaned up along side the CRDs


Tejeev:
@PhanLe1010 @ejweber Incorrect uninstallations are common to see on the support side.  Does it hurt to add these lines to the docs?  Perhaps I should add them in their own section or label the CRD section something like "Remove Longhorn Resources"

PhanLe1010:
How about can we put them into a KB article "installation trouble shooting" and add all the known conner cases there?

Tejeev:
KB articles aren't as available to the general public.  I'm not sure I follow why we would want to mention CRDs but not webhooks.  I suppose removing the CRDs should normally catch the webhooks?

PR #710, opened 2023-05-26
doc(support-bundle): concurrent generation

c3y1huang:
https://github.com/longhorn/longhorn/issues/5882


innobead (changes requested):
Please update 1.4.x doc as well, since this change is also applied to 1.4.

innobead (approved):

PR #713, opened 2023-05-31
doc(important-note): remove CRD fields

c3y1huang:
https://github.com/longhorn/longhorn/issues/3289


innobead (commented):
Users will not be able to differentiate these fields from resource spec or status, so suggest briefly explaining them via path way like xxx.spec.yyy

innobead (approved):
LGTM

PR #727, opened 2023-06-07
Add KB article outlining recent SELinux issues

ejweber:
longhorn/longhorn#5627


innobead (approved):
LGTM, especially it's good to see we even contribute a upstream fix.

PR #728, opened 2023-06-08
doc: fix broken link for selinux kb

innobead:


innobead:
cc @ejweber

ejweber:
Thanks @innobead. My previous Markdown setup was unable to resolve any of the links in our website, so I was winging it a bit here. Got a new Markdown linter and found the Hugo instructions, so this should go a bit smoother next time.

PR #732, opened 2023-06-12
fixes identation for pvc-creation for restore of sts-pvcs

Brice187:


innobead (changes requested):
Thank you. Can you also help update to other doc versions? Thanks. 

P.S. remember sign you commit.

innobead (approved):

PR #741, opened 2023-06-17
doc: remove centos from the verified list

innobead:


innobead:
@yangchiu Please remember to update the versions of OSs to reflect what version we are testing for 1.5.

cc @khushboo-rancher

PR #750, opened 2023-06-29
Update create-volumes.md

cdacosta512:
The wrong storage classes is being referenced in the pvc manifest and the replicas should drop down to 1 for single node clusters and or kept at the # of nodes the user has or else the volume will be degraded due to resources.


shuo-wu (commented):
I see. The word for the first part is misleading. The article should say that "you can use the default StorageClass `longhorn` or create a new StorageClass. You can refer to https://raw.githubusercontent.com/longhorn/longhorn/v{{< current-version >}}/examples/storageclass.yaml for the StorageClass creation".

Then the following part (pod/pvc creation) can be based on the default StorageClass. No change is needed.

PR #753, opened 2023-07-06
chore: add CIFS backupstore information

derekbit:


content/docs/1.1.3/concepts.md
```diff
@@ -263,7 +263,7 @@ Longhorn runs the sync command automatically before creating a snapshot.
 
 # 3. Backups and Secondary Storage
 
-A backup is an object in the backupstore, which is an NFS or S3 compatible object store external to the Kubernetes cluster. Backups provide a form of secondary storage so that even if your Kubernetes cluster becomes unavailable, your data can still be retrieved.
+A backup is an object in the backupstore, which is an S3, NFS or CIFS compatible object store external to the Kubernetes cluster. Backups provide a form of secondary storage so that even if your Kubernetes cluster becomes unavailable, your data can still be retrieved.
```
innobead, line 266: CIFS is only supported from 1.5, right?

innobead (changes requested):

content/docs/1.1.3/concepts.md
derekbit: Ah. You're right!

derekbit: Fixed. Thank you.

innobead (approved):

innobead:
@mantissahz You can also add the azure blob store here since it's also newly added from 1.5.0.

PR #755, opened 2023-07-11
doc: add default disk reserved percentage setting

mantissahz:
Ref: longhorn/longhorn#5958


innobead (approved):
LGTM.

@mantissahz need to remember don't forget to update the doc next time.

PR #756, opened 2023-07-12
content: Add OS kernel version recommendation in the best practice doc

shuo-wu:
When checking the cluster env of one user, I just realized that we didn't mention the kernel version recommendation. 

The recommendation is based on [the conclusion of this issue](https://github.com/longhorn/longhorn/issues/2507#issuecomment-857195496).

Notice that Longhorn v1.3.x or older version recommends Ubuntu 18.04, of which the kernel version is probably smaller than 5.8. Hence I don't add this line for them.


derekbit (dismissed):
LGTM

innobead (changes requested):
LGTM. @shuo-wu Please also update the developing versions as well like 1.5.1 and 1.6.0.

innobead:
It's fine for 1.3.x or older unsupported versions since 1.3 is going to EOL soon.

shuo-wu:
1.5.1 and 1.6.0 doc are not created yet. We can merge this PR first. Later on when launching the new version doc, there is no need to add the statement one by one.

innobead:
1.5.1 and 1.6.0 docs have been added right?

innobead (approved):

PR #762, opened 2023-08-01
doc: add a page explaining storage class parameters

james-munson:
longhorn/longhorn#4776

### DIFF
--- a/content/docs/1.5.0/references/storage-class-parameters.md
+++ b/content/docs/1.5.0/references/storage-class-parameters.md
@@ -193,9 +193,9 @@ A list of recurring jobs that are to be run on a volume.
 > More details in [Scheduling](../../volumes-and-nodes/scheduling).
 
 #### Backend Store Driver *(field: `parameters.backendStoreDriver`)*
-> Default: `""`  
+> Default: `"v1"`  
 
-  - Use "spdk" to run the v2 SPDK-based engine.
+  - Use "v2" to run the v2 SPDK-based engine (experimental in 1.5.)  If empty, "v1" is assumed.
 
 > Global setting: [V2 Data Engine](../settings#v2-data-engine).  
 > More details in [V2 Data Engine Quick Start](../../spdk/quick-start#create-a-storageclass).
--- a/content/docs/1.5.0/references/storage-class-parameters.md
+++ b/content/docs/1.5.0/references/storage-class-parameters.md
@@ -37,7 +37,6 @@ parameters:
 #  disableRevisionCounter: false
 #  replicaSoftAntiAffinity: "ignored"
 #  replicaZoneSoftAntiAffinity: "ignored"
-#  replicaDiskSoftAntiAffinity: "ignored"
 #  backendStoreDriver: ""
 ```
 
@@ -70,7 +69,7 @@ These fields will be applied for new volume creation only.  If a StorageClass is
 
 The desired number of copies (replicas) for redundancy.  
   - Must be between 1 and 20.  
-  - Replicas will be placed across the widest possible set of zones, nodes, and disks in a cluster, subject to other constraints, such as NodeSelector.
+  - Replicas will be placed across the widest possible set of zones and nodes in a cluster, subject to other constraints, such as NodeSelector.
 
 > Global setting: [Default Replica Count](../settings#default-replica-count).
 
@@ -191,15 +190,6 @@ A list of recurring jobs that are to be run on a volume.
   - Other values are "enabled" and "disabled".  
 
 > Global setting: [Replica Zone Level Soft Anti-Affinity](../settings#replica-zone-level-soft-anti-affinity).  
-> More details in [Scheduling](../../volumes-and-nodes/scheduling).
-
-#### Replica Disk Soft Anti-Affinity *(field: `parameters.replicaDiskSoftAntiAffinity`)*
-> Default: `ignored`  
-
-  - "ignored" means use the global setting.  
-  - Other values are "enabled" and "disabled".  
-
-> Global setting: TBA.
 > More details in [Scheduling](../../volumes-and-nodes/scheduling).
 
 #### Backend Store Driver *(field: `parameters.backendStoreDriver`)*
--- a/content/docs/1.6.0/advanced-resources/orphaned-data-cleanup.md
+++ b/content/docs/1.6.0/advanced-resources/orphaned-data-cleanup.md
@@ -147,4 +147,4 @@ Longhorn will not create an `orphan` resource for an orphaned directory when
   - The volume volume.meta file is missing.
 - The orphaned replica directory is on an evicted node.
 - The orphaned replica directory is in an evicted disk.
-- The orphaned data cleanup mechanism does not clean up a stable replica, also known as an error replica. Instead, the stale replica is cleaned up according to the [staleReplicaTimeout](../../volumes-and-nodes/create-volumes/#creating-longhorn-volumes-with-kubectl) setting.
+- The orphaned data cleanup mechanism does not clean up a stale replica, also known as an error replica. Instead, the stale replica is cleaned up according to the [staleReplicaTimeout](../../volumes-and-nodes/create-volumes/#creating-longhorn-volumes-with-kubectl) setting.
--- a/content/docs/1.6.0/references/settings.md
+++ b/content/docs/1.6.0/references/settings.md
@@ -577,7 +577,10 @@ When this setting is un-checked, the Longhorn Manager will schedule replicas on 
 
 When this setting is checked, the Longhorn Manager will allow scheduling on nodes with existing healthy replicas of the same volume.
 
-When this setting is un-checked, the Longhorn Manager will not allow scheduling on nodes with existing healthy replicas of the same volume.
+When this setting is un-checked, Longhorn Manager will forbid scheduling on nodes with existing healthy replicas of the same volume.
+
+> **Note:**
+>   - This setting is superseded if replicas are forbidden to share a zone by the Replica Zone Level Anti-Affinity setting.
 
 #### Replica Zone Level Soft Anti-Affinity
 
@@ -585,7 +588,7 @@ When this setting is un-checked, the Longhorn Manager will not allow scheduling 
 
 When this setting is checked, the Longhorn Manager will allow scheduling new replicas of a volume to the nodes in the same zone as existing healthy replicas.
 
-When this setting is un-checked, Longhorn Manager will not allow scheduling new replicas of a volume to the nodes in the same zone as existing healthy replicas.
+When this setting is un-checked, Longhorn Manager will forbid scheduling new replicas of a volume to the nodes in the same zone as existing healthy replicas.
 
 > **Note:**
 >   - Nodes that don't belong to any zone will be treated as if they belong to the same zone.
@@ -597,10 +600,11 @@ When this setting is un-checked, Longhorn Manager will not allow scheduling new 
 
 When this setting is checked, the Longhorn Manager will allow scheduling new replicas of a volume to the same disks as existing healthy replicas.
 
-When this setting is un-checked, Longhorn Manager will not allow scheduling new replicas of a volume to the same disks as existing healthy replicas.
+When this setting is un-checked, Longhorn Manager will forbid scheduling new replicas of a volume to the same disks as existing healthy replicas.
 
 > **Note:**
 >   - Even if the setting is "true" and disk sharing is allowed, Longhorn will seek to use a different disk if possible, even if on the same node.
+>   - This setting is superseded if replicas are forbidden to share a zone or a node by either of the other Soft Anti-Affinity settings.
 
 #### Replica Auto Balance
 
--- a/content/docs/1.6.0/references/settings.md
+++ b/content/docs/1.6.0/references/settings.md
@@ -60,6 +60,7 @@ weight: 1
   - [Disable Scheduling On Cordoned Node](#disable-scheduling-on-cordoned-node)
   - [Replica Node Level Soft Anti-Affinity](#replica-node-level-soft-anti-affinity)
   - [Replica Zone Level Soft Anti-Affinity](#replica-zone-level-soft-anti-affinity)
+  - [Replica Disk Level Soft Anti-Affinity](#replica-disk-level-soft-anti-affinity)
   - [Replica Auto Balance](#replica-auto-balance)
   - [Storage Minimal Available Percentage](#storage-minimal-available-percentage)
   - [Storage Over Provisioning Percentage](#storage-over-provisioning-percentage)
@@ -295,6 +296,7 @@ This information will help us gain insights how Longhorn is being used, which wi
     - Replica Replenishment Wait Interval
     - Replica Soft Anti Affinity
     - Replica Zone Soft Anti Affinity
+    - Replica Disk Soft Anti Affinity
     - Restore Concurrent Limit
     - Restore Volume Recurring Jobs
     - Snapshot Data Integrity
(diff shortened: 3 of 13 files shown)


content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+## Overview
+
+Storage Class as a resource object has a number of settable parameters.  See also [Kubernetes Storage Class](https://kubernetes.io/docs/concepts/storage/storage-classes).  
+Most of the ones in this list are specific to Longhorn's driver.  A few are common to all Kubernetes storage classes:
+
+### Default Class
+Set this Storage Class as the default for dynamic (Kubernetes-driven) volume creation.  
+Type: Boolean  
+Notes: This should only be true for one StorageClass object.  If not, either all volume creations will fail, or else one of the competing defaults will be chosen, and it may be hard to predict which one.  
+***(Which is it?  Give this a try.)***
```
ejweber, line 15: I'm sure you are planning on cleaning this up. Commenting here so we don't forget.

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+title: Storage Class Settings
+weight: 3
+---
+
+## Overview
+
+Storage Class as a resource object has a number of settable parameters.  See also [Kubernetes Storage Class](https://kubernetes.io/docs/concepts/storage/storage-classes).  
+Most of the ones in this list are specific to Longhorn's driver.  A few are common to all Kubernetes storage classes:
+
+### Default Class
```
ejweber, line 11: If I am not mistaken, whether a StorageClass is default or not is controlled by a label, not a field. We should make this explicit or (and this may actually be my preference) simply not document it since the Kubernetes docs already handle it fine.

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+### Provisioner
+Specifies the plugin that will be used for dynamic creation of persistent volumes.  For Longhorn, that is always "driver.longhorn.io"
+
+### Reclaim Policy
+Define what happens to a persistent volume when released from its claim (PVC), when the claim is deleted, say.  This is strictly a Kubernetes setting.  Longhorn itself does not validate the setting nor make any decisions based on it.  
+Type:  Enumeration, either "Delete" or "Retain".  
+Notes:  This is a two-valued enumeration rather than a Boolean because there was a third setting, "Recycle", which is deprecated and should be avoided.
+
+## Longhorn-specific Parameters
+Note that some of these parameters also exist and may be specified in global settings.  Which one is used in a specific instance will depend on whether a volume is created via Kubernetes or the UI.
```
ejweber, line 26: Since this document is specifically about StorageClass parameters, I suggest:

```suggestion
Note that some of these parameters also exist and may be specified in global settings.  When a volume is provisioned with Kubernetes against a particular StorageClass, StorageClass parameters override these global settings.
```

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
ejweber, line 1: In general, I suggest we move this to https://longhorn.io/docs/1.6.0/references/storage-class-settings and make it like https://longhorn.io/docs/1.6.0/references/settings/ as much as possible. (Its content feels very similar to the content of that page.)

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Provisioner
+Specifies the plugin that will be used for dynamic creation of persistent volumes.  For Longhorn, that is always "driver.longhorn.io"
+
+### Reclaim Policy
+Define what happens to a persistent volume when released from its claim (PVC), when the claim is deleted, say.  This is strictly a Kubernetes setting.  Longhorn itself does not validate the setting nor make any decisions based on it.  
+Type:  Enumeration, either "Delete" or "Retain".  
+Notes:  This is a two-valued enumeration rather than a Boolean because there was a third setting, "Recycle", which is deprecated and should be avoided.
+
+## Longhorn-specific Parameters
```
ejweber, line 25: Should we clarify that the previous settings were appropriate top level, but that these exist within the parameters field?

```
topLevelSetting:
topLevelSetting:
parameters:
  longhornSetting:
```

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+Note that some of these parameters also exist and may be specified in global settings.  Which one is used in a specific instance will depend on whether a volume is created via Kubernetes or the UI.
+
+### Replica Count
```
ejweber, line 28: I'm thinking we need to be careful about the name we use here. For example, the actual storage class setting we recognize is `numberOfReplicas`, which does not appear in this text.

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
ejweber, line 1: I'm thinking a lot of these setting references could use links to other places in the documentation. Maybe this would limit the amount of extra information we have to maintain here?

ejweber: There are a number of parameters not yet covered by this document. In retrospect, I'm thinking I am reviewing an early draft? If so, sorry to drop so many comments on you!

https://github.com/longhorn/longhorn/blob/19e8fefd3ace7fb66c3f3521fc471b60a829b155/examples/storageclass.yaml#L9-L23

content/docs/1.6.0/volumes-and-nodes/storage-class-parms.md
```diff
@@ -0,0 +1,95 @@
...
+  Type:  String with JSON map. e.g., '{\"url\":\"https://backing-image-example.s3-region.amazonaws.com/test-backing-image\"}'.  
+  Notes:
+  - If the backing image name is not specified, setting this field is meaningless.
+  - If the backing image does not exist, Longhorn will use this field to create one.  Otherwise, Longhorn will use it to verify the selected backing image.
+  - Be careful of the quotes here.
+
+### Remove Snapshots During Filesystem Trim
+Allow automatically removing snapshots during filesystem trim for this Longhorn storage class.  
+Type: Enumerated string, one of "ignored" (default), "enabled", or "disabled".  
+Notes:  This seems like it overlaps the global setting `Remove Snapshots During Filesystem Trim`, and it does.  The idea is the same, but one key difference is that the global is a Boolean, and this one an enumeration.  This one sets a policy for all volumes created with this storage class.  The setting "enabled" means always allow it, "disabled" means never allow it, and "ignored" (the default) means the global setting should be used.   See more in [Trim Filesystem](../trim-filesystem).
```
shuo-wu, line 90: You can add a refer link to the setting. Then there is no need to explain how this field works.

shuo-wu (commented):
1. It's better to add one example at the beginning.
2. I think we can explain how Longhorn creates the default SC. And why we use a ConfigMap to create that default SC (since some fields are immutable.)

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+  numberOfReplicas: "3"
+  staleReplicaTimeout: "2880"
+  fromBackup: ""
+  fsType: "ext4"
+  diskSelector: "ssd,fast"
+  nodeSelector: "storage,fast"
+#  recurringJobSelector: '[{"name":"snap-group", "isGroup":true},
+#                          {"name":"backup", "isGroup":false}]'
+```
+
```
shuo-wu, line 28: Add line `## Built-in Fields`

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+> Default: `true`
+
+#### Reclaim Policy
+> Default: `Delete`
+
+#### Volume Binding Mode
+> Default `Immediate`
+
+## Longhorn-specific Parameters
+Note that some of these parameters also exist and may be specified in global settings.  When a volume is provisioned with Kubernetes against a particular StorageClass, StorageClass parameters override the global settings.
```
shuo-wu, line 45: After the recent discussion about the StorageClass parameters, I prefer to add the below note:
```
These fields will be applied for volume creation only. Neither Longhorn nor Kubernetes is responsible for keeping the current SC parameters with that of the volumes previously created by this SC in sync.
```

shuo-wu (dismissed):
In general LGTM

ejweber (commented):
Much improved. A couple of questions still. One big thing we need to do is "backport" this to other versions of the website. Probably we don't need to go all the way back to the beginning of time, but to the versions that aren't EOL. That will be annoying, as some of these options are new.

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+volumeBindingMode: Immediate
+parameters:
+  numberOfReplicas: "3"
+  staleReplicaTimeout: "2880"
+  fromBackup: ""
+  fsType: "ext4"
+  diskSelector: "ssd,fast"
+  nodeSelector: "storage,fast"
+#  recurringJobSelector: '[{"name":"snap-group", "isGroup":true},
+#                          {"name":"backup", "isGroup":false}]'
```
innobead, line 26: Add all parameters here, but  comment out the optional parameters as below. However, you can give the optional parameter with the default value. 

```yaml
    parameters:
      numberOfReplicas: "{{ .Values.persistence.defaultClassReplicaCount }}"
      staleReplicaTimeout: "30"
      fromBackup: ""
      # fsType: ".."
      # mkfsParams: ".."
      # migratable: ".."
      ...
```

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+Some fields are common to all Kubernetes storage classes.
+See also [Kubernetes Storage Class](https://kubernetes.io/docs/concepts/storage/storage-classes).  
+
+#### Provisioner
```
innobead, line 32: Suggest have a mapping for the field name like

```
#### Provisioner (field: `provisioner`)
```

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+Specifies the plugin that will be used for dynamic creation of persistent volumes.  For Longhorn, that is always "driver.longhorn.io"
+
+#### Allow Volume Expansion
+> Default: `true`
+
+#### Reclaim Policy
```
innobead, line 38: Can add a upstream doc for reference, the same for other builtin field.

content/docs/1.6.0/references/storage-class-parameters.md
innobead: This is a good addition.

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+> Default: `Delete`
+
+#### Volume Binding Mode
+> Default `Immediate`
+
+## Longhorn-specific Parameters
+Note that some of these parameters also exist and may be specified in global settings.  When a volume is provisioned with Kubernetes against a particular StorageClass, StorageClass parameters override the global settings.
+
+#### Number Of Replicas
```
innobead, line 47: Suggest changing to below. Please also update other parameters.

```
### Number of Rreplicas (filed: `parameters.numberOfReplicas`)
```

innobead (changes requested):
In general, LGTM. Just some minor suggestions. 

Good job, James.

content/docs/1.6.0/advanced-resources/orphaned-data-cleanup.md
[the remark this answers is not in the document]
innobead: https://github.com/longhorn/website/pull/762#discussion_r1289818683

PhanLe1010:
The PR LGTM once all the above review comments are resolved.

content/docs/1.6.0/advanced-resources/orphaned-data-cleanup.md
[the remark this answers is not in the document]
ejweber: Field name was added as suggested in https://github.com/longhorn/website/pull/762#discussion_r1289818683. Resolving conversation.

content/docs/1.6.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,154 @@
...
+  nodeSelector: "storage,fast"
+#  recurringJobSelector: '[{"name":"snap-group", "isGroup":true},
+#                          {"name":"backup", "isGroup":false}]'
+```
+
+Some fields are common to all Kubernetes storage classes.
+See also [Kubernetes Storage Class](https://kubernetes.io/docs/concepts/storage/storage-classes).  
+
+#### Provisioner
+Specifies the plugin that will be used for dynamic creation of persistent volumes.  For Longhorn, that is always "driver.longhorn.io"
```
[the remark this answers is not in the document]
ejweber: This was fixed. Resolving conversation.

ejweber (changes requested):
It looks like you addressed every comment of mine but one. This is looking really good. Still need to add a version of this (minus newly introduced parameters) to other versions IMO. Let me know if you disagree.

innobead (changes requested):
LGTM. The only remaining from my side is to have this page for the latest existing patch or upcoming patch release doc as well, such as 1.3.3. 1.4.3 & 1.4.4, 1.5.1 & 1.5.2.

P.S. No need to have that in 1.3.4, since it's EOL and we will not release it.

james-munson:
I added the 1.4.x and 1.5.x versions.  
1.4.x are all identical.
1.5.x are all identical to each other, and differ from 1.4.x in that the backendStoreDriver (v2 engine) parameter is added.
1.6.0 differs from 1.5.x in that the replica **disk** soft anti-affinity is added.  (Looking ahead to when Eric's PR is merged!  It will need one tiny tweak when that happens, to point to the global setting, which doesn't exist yet, and gives an invalid reference in the build.)

james-munson:
Also added documentation for "replica disk soft anti-affinity" to global settings and replica scheduling section.

ejweber (dismissed):
Thanks for the Replica Disk Soft Anti Affinity addition. Looks good!

I am personally fine with not including this page on the v1.3.x branch given its EOL status, so approving. Not sure where others will fall.

PhanLe1010:
I am fine with not having it in 1.3.x too

innobead (changes requested):
LGTM, but some comments need to resolve first.

content/docs/1.6.0/references/settings.md
```diff
@@ -589,6 +591,17 @@ When this setting is un-checked, Longhorn Manager will not allow scheduling new
...
 
+#### Replica Disk Level Soft Anti-Affinity
+
+> Default: `true`
+
+When this setting is checked, the Longhorn Manager will allow scheduling new replicas of a volume to the same disks as existing healthy replicas.
+
+When this setting is un-checked, Longhorn Manager will not allow scheduling new replicas of a volume to the same disks as existing healthy replicas.
+
+> **Note:**
```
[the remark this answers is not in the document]
innobead: oh, it was reviewed. It's about backward compatibility. https://github.com/longhorn/longhorn/blob/master/enhancements/20230718-disk-anti-affinity.md#upgrade-strategy.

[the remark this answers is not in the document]
james-munson: Added, thanks.

content/docs/1.5.0/references/storage-class-parameters.md
```diff
@@ -0,0 +1,206 @@
...
+#### Replica Zone Soft Anti-Affinity *(field: `parameters.replicaZoneSoftAntiAffinity`)*
+> Default: `ignored`  
+
+  - "ignored" means use the global setting.  
+  - Other values are "enabled" and "disabled".  
+
+> Global setting: [Replica Zone Level Soft Anti-Affinity](../settings#replica-zone-level-soft-anti-affinity).  
+> More details in [Scheduling](../../volumes-and-nodes/scheduling).
+
+#### Backend Store Driver *(field: `parameters.backendStoreDriver`)*
```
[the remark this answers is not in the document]
james-munson: Good catch.  Fixed, thanks.

innobead (approved):

PR #763, opened 2023-08-02
doc(telemetric): collect Snapshot Data Integrity CronJob setting as boolean

c3y1huang:
Backported to v1.5.2.

ref: https://github.com/longhorn/longhorn/issues/6438, https://github.com/longhorn/longhorn/issues/6410


PhanLe1010 (approved):
LGTM

PR #770, opened 2023-08-30
doc(scheduling): add allow empty selector settings

ChanYiLin:
ref: [longhorn/longhorn#4826](https://github.com/longhorn/longhorn/issues/4826)

### DIFF
--- a/content/docs/1.6.0/references/settings.md
+++ b/content/docs/1.6.0/references/settings.md
@@ -64,6 +64,8 @@ weight: 1
   - [Storage Minimal Available Percentage](#storage-minimal-available-percentage)
   - [Storage Over Provisioning Percentage](#storage-over-provisioning-percentage)
   - [Storage Reserved Percentage For Default Disk](#storage-reserved-percentage-for-default-disk)
+  - [Allow Empty Node Selector Volume](#allow-empty-node-selector-volume)
+  - [Allow Empty Disk Selector Volume](#allow-empty-disk-selector-volume)
 - [Danger Zone](#danger-zone)
   - [Concurrent Replica Rebuild Per Node Limit](#concurrent-replica-rebuild-per-node-limit)
   - [Kubernetes Taint Toleration](#kubernetes-taint-toleration)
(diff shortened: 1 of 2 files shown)


derekbit (dismissed):

james-munson:
This is only a global setting, not a storage class or volume parameter?

ChanYiLin:
Hi @james-munson
Yes, it is a global setting

The use case is for example:
User has a set of hosts with SSD and have the tag: ["SSD"]
But user just want Volumes with selector: ["SSD"] to be scheduled on those hosts.
Without this setting, user needs to add selectors for all other Volumes with different selectors or Volumes with empty selector will still have chance to be placed on those hosts. This is inefficient.

We don't need this setting to be Volume level because it can be achieved by using selectors/tags already
Lets say
For a `nodeA:[“ssd”]`
You want `VolumeA:[]`  to be able to be scheduled on it and `VolumeB:[]` cant be scheduled on it
You can just give `VolumeA: ["SSD"]` label to achieve it.

ejweber (approved):

roger-ryao:
Hi @ChanYiLin 
   https://github.com/longhorn/longhorn/issues/4826  is set for milestone `v1.6.0`. I believe that https://github.com/longhorn/website/pull/770 should be placed in `/content/docs/1.6.0` instead of `/content/docs/1.5.1`, is that correct?

ChanYiLin:
> Hi @ChanYiLin [longhorn/longhorn#4826](https://github.com/longhorn/longhorn/issues/4826) is set for milestone `v1.6.0`. I believe that #770 should be placed in `/content/docs/1.6.0` instead of `/content/docs/1.5.1`, is that correct?

Yes, you are right, thanks!

roger-ryao (changes requested):
It should be placed in `/content/docs/1.6.0` instead of `/content/docs/1.5.1`

ChanYiLin:
@roger-ryao fixed, thanks!

innobead (approved):

PR #771, opened 2023-09-04
doc(upgrade): modify upgrade path example

mantissahz:
Modify the upgrade path supported matrix to use `current-version` for 1.5.x and 1.6.x upgrade path documents.

### DIFF
--- a/content/docs/1.5.0/deploy/upgrade/_index.md
+++ b/content/docs/1.5.0/deploy/upgrade/_index.md
@@ -13,7 +13,7 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 Since Longhorn v1.5.0, Longhorn only allows upgrades from supported versions. If upgrading from any unsupported version, the upgrade will fail. However, users can revert to the previous state without any service interruption or downtime.
 
-In addition, Longhorn disallows downgrading to any previous version to prevent unexpected system statuses caused by potential function incompatibility, deprecation, or removal. Please refer to the following matrix to understand the supported upgrade versions:
+In addition, Longhorn disallows downgrading to any previous version to prevent unexpected system statuses caused by potential function incompatibility, deprecation, or removal. Please refer to the following matrix to understand the supported upgrade versions and there is an example for each upgrade path:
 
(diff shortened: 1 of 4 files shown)


innobead (changes requested):
This is just an example, so probably we don't need to change for each release. It seems extra cost to maintain this page over time. WDYT?

P.S. We need to emphasize this matrix is just an example.

mantissahz:
>This is just an example, so probably we don't need to change for each release. It seems extra cost to maintain this page over time. WDYT?
P.S. We need to emphasize this matrix is just an example.

@innobead Agree with you and modify the documents (1.5.x and 1.6.0) only to change `v1.5.0` to `v1.5.1` for examples in the matrix. and then we don't need to modified it in the future.

content/docs/1.5.0/deploy/upgrade/_index.md
```diff
@@ -13,7 +13,7 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 Since Longhorn v1.5.0, Longhorn only allows upgrades from supported versions. If upgrading from any unsupported version, the upgrade will fail. However, users can revert to the previous state without any service interruption or downtime.
 
-In addition, Longhorn disallows downgrading to any previous version to prevent unexpected system statuses caused by potential function incompatibility, deprecation, or removal. Please refer to the following matrix to understand the supported upgrade versions:
+In addition, Longhorn disallows downgrading to any previous version to prevent unexpected system statuses caused by potential function incompatibility, deprecation, or removal. Please refer to the following matrix to understand the supported upgrade versions there is an example for each upgrade path:
```
innobead, line 16: the change is missing `and` if compared with other doc changes.

innobead (changes requested):
LGTM, just one comment needs to resolve.

content/docs/1.5.0/deploy/upgrade/_index.md
mantissahz: Modified. Thanks.

innobead (approved):

PR #784, opened 2023-10-06
Add KB troubleshooting filesystem trim error 

PhanLe1010:
longhorn/longhorn#6854

### DIFF
--- a/content/kb/troubleshooting-fstrim-error.md
+++ b/content/kb/troubleshooting-fstrim-error.md
@@ -1,0 +1,74 @@
+---
+title: "Troubleshooting: fstrim doesn't work on old kernel"
+author: Phan Le
+draft: false
+date: 2023-10-06
+catelogies:
+  - "trim"
+---
+
+## Applicable versions
+
+* Longhorn version >= `v1.4.0`
+* Node kernel version <= `v4.12`
+
+## Symptoms
+
+When running filesystem trim (either by Longhorn UI or manually by `fstrim` command on the host), it hits the error similar to:
+
+```
+unable to trim filesystem for volume pvc-e381424a-4866-447a-a75c-3096036f7846: cannot find volume pvc-e381424a-4866-447a-a75c-3096036f7846 mount info on host: failed to execute: nsenter [--mount=/host/proc/20357/ns/mnt --net=/host/proc/20357/ns/net fstrim /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount], output , stderr fstrim: /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount: FITRIM ioctl failed: Input/output error\n: exit status 255
+```
+or
+```
+unable to trim filesystem for volume pvc-e381424a-4866-447a-a75c-3096036f7846: cannot find volume pvc-e381424a-4866-447a-a75c-3096036f7846 mount info on host: failed to execute: nsenter [--mount=/host/proc/20357/ns/mnt --net=/host/proc/20357/ns/net fstrim /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount], output , stderr fstrim: /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount: the discard operation is not supported : exit status 1
+```
+
+## Details
+
+This is caused by a problem in old kernel SCSI driver which sets the wrong provisioning mode for the SCSI device that advertised
+both UNMAP and WRITE SAME capability as detailed in this kernel patch https://github.com/torvalds/linux/commit/bcd069bb250acf6088b60d189ab3ec3ae8dd11a5
+
+### Troubleshooting steps
+
+#### Case 1: Volume is using old engine image or running in the old instance manager pod
+1. Double-check the engine image of the volume is >= `v1.4.0` in the volume detail page in Longhorn UI. Only engine
+image >= `v1.4.0` supports filesystem trim feature.
+1. If the volume's engine's image was recently live upgraded from a version < `v1.4.0` to >= `v1.4.0`, make sure that the
+volume is detached/reattached first to activate the trim feature in Longhorn volume. You can do this by manually scale
+down and up the workload deployment that is using the volume. See more details at https://longhorn.io/docs/1.4.0/volumes-and-nodes/trim-filesystem/#prerequisites
+1. After the above steps, try to run filesystem trim again to see if the problem still exists. If the problem still exists,
+move on to the next case below.
+
+#### Case 2: Old node kernel version
+1. Find the node that the volume is currently attached to and ssh into that node
+1. Find the major and minor of a Longhorn block device by `ls -l /dev/longhorn/<volume-name>`. For example:
+    ```bash
+    [root@phan-v147-cloudera-pool1-f1dec634-cm59v ~]# ls -l /dev/longhorn/testvol
+    brw-rw----. 1 root root 8, 0 Oct  6 20:24 /dev/longhorn/testvol
+    ```
+    In this case the major:minor versions is `8:0`
+1. Find the corresponding SCSI device's address by `lsscsi -d`. For example,
+    ```bash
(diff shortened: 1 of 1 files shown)


innobead (approved):
LGTM.

@jillian-maroket please help review this.

ejweber (dismissed):
Read through for my own understanding. LGTM!

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,71 @@
...
+
+This is caused by a problem in old kernel SCSI driver which sets the wrong provisioning mode for the SCSI device that supports
+both UNMAP and WRITE SAME capability as detailed in this kernel patch https://github.com/torvalds/linux/commit/bcd069bb250acf6088b60d189ab3ec3ae8dd11a5
+
+### Troubleshooting steps
+
+1. Double-check the engine image of the volume is >= `v1.4.0` in the volume detail page in Longhorn UI. Only engine
+image >= `v1.4.0` supports filesystem trim feature.
+1. If the volume's engine's image was recently live upgraded from a version < `v1.4.0` to >= `v1.4.0`, make sure that the
+volume is detached/reattached first to activate the trim feature in Longhorn volume. You can do this by manually scale
```
shuo-wu, line 37: NIT: Do we need to separate this fstrim failure into 2 cases:
1. Engines are using old IM pod/engine image
2. The node kernel version

shuo-wu (dismissed):
In general LGTM

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,71 @@
...
+```
+or
+```
+unable to trim filesystem for volume pvc-e381424a-4866-447a-a75c-3096036f7846: cannot find volume pvc-e381424a-4866-447a-a75c-3096036f7846 mount info on host: failed to execute: nsenter [--mount=/host/proc/20357/ns/mnt --net=/host/proc/20357/ns/net fstrim /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount], output , stderr fstrim: /var/lib/kubelet/plugins/kubernetes.io/csi/driver.longhorn.io/d2df1133f3440486ddec39370380eeed7a3c71499981d63fc80e43c7ca9f4c9e/globalmount: the discard operation is not supported : exit status 1
+```
+
+## Details
+
+This is caused by a problem in old kernel SCSI driver which sets the wrong provisioning mode for the SCSI device that supports
```
derekbit, line 29: FYI
The legacy way in an old kernel is using write_same, but unfortunately, longhorn tgt doesn't implement the operation ([codes](https://github.com/rancher/tgt/blob/longhorn_oct2019/usr/bs_longhorn.c)). That's why the kernel falls back to `disabled`.

derekbit (approved):

content/kb/troubleshooting-fstrim-error.md
PhanLe1010: Thanks for the link to the code!

Even though our engine doesn't support write_same, the created block device somehow still advertises that capability:
```
root@phan-v406-dt-pool2-e5bcaaa3-bxtvt:~# sg_vpd /dev/longhorn/testvol -p lbpv
Logical block provisioning VPD page (SBC):
  Unmap command supported (LBPU): 1
  Write same (16) with unmap bit supported (LBPWS): 1
  Write same (10) with unmap bit supported (LBPWS10): 1
  Logical block provisioning read zeros (LBPRZ): 1
  Anchored LBAs supported (ANC_SUP): 0
  Threshold exponent: 0 [threshold sets not supported]
  Descriptor present (DP): 0
  Minimum percentage: 0 [not reported]
  Provisioning type: 2 (thin provisioned)
  Threshold percentage: 0 [percentages not supported]
```

From old kernel perspective, it choose write_same over unmap which cause the problem. However, I am not sure how to modify LUN to disable advertising write_same capability. The only option I see it Thin-provisioning. When enable Thin-provisioning.

content/kb/troubleshooting-fstrim-error.md
PhanLe1010: Quick question, how the instance manager pod can affect trim?

content/kb/troubleshooting-fstrim-error.md
derekbit: I think it is due to that longhorn doesn't return `-1` to the upper caller for the unsupported write_same operation.
https://github.com/rancher/tgt/blob/longhorn_oct2019/usr/bs_longhorn.c#L199

content/kb/troubleshooting-fstrim-error.md
shuo-wu: For example, an old volume was running in a v1.3.x instance manager pod then got upgraded to engine image v1.4.x. In this case, the upgraded engine was still in the old IM pod, of which the tgt does not support `unmap`.

PhanLe1010: I see. Good point. I separated into 2 cases

content/kb/troubleshooting-fstrim-error.md
PhanLe1010: I am not sure if changing the return value over there can disable advertising the `write_same` capability because in the older Longhorn version (< v1.4.0), we don't return -1 but Longhorn volume still does NOT advertise the `write_same` capability. 

It is when we change `Thin-provisioning` yo `true` in Longhorn > v1.4.0, both unmap and write_same capabilitie are advertised  

Opening a ticket for further investigation https://github.com/longhorn/longhorn/issues/6876

james-munson (approved):

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,74 @@
...
+    ```
+    In this case the major:minor versions is `8:0`
+1. Find the corresponding SCSI device's address by `lsscsi -d`. For example,
+    ```bash
+    [root@phan-v147-cloudera-pool1-f1dec634-cm59v ~]# lsscsi -d
+    [3:0:0:0]    storage IET      Controller       0001  -
+    [3:0:0:1]    disk    IET      VIRTUAL-DISK     0001  /dev/sda [8:0]
+    ```
+    In this case, the corresponding corresponding SCSI device's address is `[3:0:0:1]` because it also has major:minor versions as `8:0`
+1. Find the provisioning mode of SCSI device by `find /sys/ -name provisioning_mode -exec grep -H . {} + | sort`. For example:
```
jillian-maroket, line 58:
```suggestion
1. Identify the provisioning mode of the SCSI device using the command `find /sys/ -name provisioning_mode -exec grep -H . {} + | sort`. 

Example:
```

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,74 @@
...
+    ```bash
+    [root@phan-v147-cloudera-pool1-f1dec634-cm59v ~]# find /sys/ -name provisioning_mode -exec grep -H . {} + | sort
+    /sys/devices/platform/host3/session1/target3:0:0/3:0:0:1/scsi_disk/3:0:0:1/provisioning_mode:writesame_16
+    ```
+    In this case, looks at the device with the address `3:0:0:1`, it has provisioning mode as `writesame_16` (in some cases,
+    it can be `disable`) but not the correct value `unmap`
+1. Check the kernel version by `uname -r`. If the kernel version is < `v4.12`, you have hit this issue
```
jillian-maroket, line 65:
```suggestion
1. Identify the kernel version using the command `uname -r`. 
```

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Solution
+
+Upgrade the kernel to a recommended version as in the best practice page https://longhorn.io/docs/1.5.1/best-practices/#operating-system.
```
jillian-maroket, line 69:
```suggestion
1. If the kernel version is earlier than `v4.12`, upgrade the kernel.
```

content/kb/troubleshooting-fstrim-error.md
```diff
@@ -0,0 +1,74 @@
...
+At the time when writing this article, the recommended kernel version is >= `v5.8`
```
jillian-maroket, line 70:
```suggestion
At the time of writing, Longhorn recommends upgrading to `v5.8` or later. For more information about verified distributions and versions, see [Best Practices](https://longhorn.io/docs/1.5.1/best-practices/#operating-system).
```

content/kb/troubleshooting-fstrim-error.md
jillian-maroket: The solution is part of Case 2 so we can consider it the final step in that procedure.

content/kb/troubleshooting-fstrim-error.md
jillian-maroket: I moved the kernel version info to the next step.

jillian-maroket (commented):
@PhanLe1010 I understand that your commit has been merged but I still reviewed the content for clarity and consistency. Let me know if the changes work.

PR #785, opened 2023-10-09
Update URL for NFS backupstore example file

devenkulkarni:


james-munson (changes requested):
As @derekbit points out in the base issue, https://github.com/longhorn/longhorn/issues/6867, references in the archived docs for old versions should point to the archived tree, not to head of master.

PR #796, opened 2023-11-11
doc(metrics): add pvc namespace to volume metrics

antoninferrand:
https://github.com/longhorn/longhorn/issues/7077


innobead:
@antoninferrand please update 1.5.3 as well. Thanks!

PR #847, opened 2024-01-16
Improve 'Upgrade Path Enforcement' content

jillian-maroket:
I reorganized and reworded the existing content, and added details from [this Slack thread](https://suse.slack.com/archives/C02HBK5AVC5/p1703730572042269).

My initial commit only covers v1.5.0 and v1.6.0, but I can expand the scope (if necessary) after I receive feedback.

### DIFF
--- a/content/docs/1.5.0/deploy/upgrade/_index.md
+++ b/content/docs/1.5.0/deploy/upgrade/_index.md
@@ -11,11 +11,11 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 # Upgrade Path Enforcement
 
-Since Longhorn v1.5.0, Longhorn only allows upgrades from supported versions. If upgrading from any unsupported version, the upgrade will fail. However, users can revert to the previous state without any service interruption or downtime.
+Starting with v1.5.0, Longhorn only allows upgrades from supported versions. When you attempt to upgrade from an unsupported version, the upgrade automatically fails but you can revert to the previously installed version without any service interruption or downtime.
 
-In addition, Longhorn disallows downgrading to any previous version to prevent unexpected system statuses caused by potential function incompatibility, deprecation, or removal. Please refer to the following matrix to understand the supported upgrade versions and there is an example for each upgrade path:
+The following table outlines the supported upgrade paths.
 
-  |  Current version |  Upgrading version |  Allow | Example |
+  |  Current version |  Target version |  Allowed | Example |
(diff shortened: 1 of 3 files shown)


content/docs/1.5.0/deploy/upgrade/_index.md
```diff
@@ -25,10 +25,10 @@ In addition, Longhorn disallows downgrading to any previous version to prevent u
   |  x.y.*      |  x.(y-1).*  |   X    |  v1.6.0  to  v1.5.0  |
   |  x.y.*      |  x.y.(*-1)  |   X    |  v1.5.1  to  v1.5.0  |
 
-[^lastMinorVersion]: Longhorn only allows upgrades from any patch version of the last minor release before the new major version. (For example, v1.30.* is allowed to upgrade to v2.0.*, given that v1.30 is the last minor release branch before 2.0.)
+[^lastMinorVersion]: Longhorn only allows upgrades from a patch version of the last minor release before the new major version. For example, you can upgrade from v1.3.* to v2.0.* because v1.3.0 is the last minor release before v2.0.
 
-> **Warning**:
-> * Upgrade path enforcement is introduced in Longhorn v1.5.0, which means that downgrading from v1.5.0 to any previous version is possible. **Please note that downgrading is not supported**.
+> **Important:**
+> If necessary, you can downgrade from v1.5.0 to an earlier version.
```
innobead, line 31: The `upgrade enforcement & downgrade prevention` has been introduced since 1.5.0. If users try to downgrade to any previous version < 1.5, it can't be prevented as the capability was not implemented in older versions. 

The revised sentence may give the impression that downgrading is allowed and supported, but in reality, it is not. Even though it can be carried out as described above, we must emphasize that downgrading is unsupported and we do not recommend users to do so.

content/docs/1.5.0/deploy/upgrade/_index.md
```diff
@@ -11,11 +11,11 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 # Upgrade Path Enforcement
```
innobead, line 12: Please also update the title to `Upgrade Path Enforcement and Downgrad Prevention`

innobead (changes requested):

jillian-maroket:
@innobead I updated the descriptions based on your feedback. Note that I placed the warning before the table to improve the content flow.

innobead (approved):
Good upate.

PR #855, opened 2024-01-18
Add aliases metadata to files in v2-data-engine

jillian-maroket:
Google search results for topics related to the V2 Data Engine point to `longhorn.io > docs > latest > spdk`. However, the links are broken because the folder was changed to 'v2-data-engine'.

We can use [aliases](https://gohugo.io/content-management/urls/#aliases) to redirect to the new URLs. I added the metadata to the affected v1.5.3 and v1.6.0 files.

### DIFF
--- a/content/docs/1.5.3/v2-data-engine/features.md
+++ b/content/docs/1.5.3/v2-data-engine/features.md
@@ -1,6 +1,8 @@
 ---
 title: Features
 weight: 1
+aliases:
+- /spdk/features.md
 ---
 
 - Volume lifecycle (creation, attachment, detachment and deletion)
--- a/content/docs/1.6.0/v2-data-engine/features/automatic-offline-replica-rebuilding.md
+++ b/content/docs/1.6.0/v2-data-engine/features/automatic-offline-replica-rebuilding.md
@@ -2,7 +2,7 @@
 title: Automatic Offline Replica Rebuilding
 weight: 10
 aliases:
-- /spdk/features/automatic-offline-replica-rebuilding.md
+- /spdk/automatic-offline-replica-rebuilding.md
 ---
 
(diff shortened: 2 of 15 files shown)


content/docs/1.5.3/v2-data-engine/features.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Features
 weight: 1
+aliases:
+- /spdk/features.md
```
innobead, line 5: It seems https://deploy-preview-855--longhornio.netlify.app/docs/1.5.3/spdk/features not working?

innobead (changes requested):

content/docs/1.5.3/v2-data-engine/features.md
jillian-maroket: I assume that the redirects will work after the aliases are actually published. Since this PR has not been merged and the 'spdk' folder no longer exists, any links containing 'spdk' will be broken.

innobead: Ok, we can see how it will work after the merge.

content/docs/1.6.0/v2-data-engine/features/automatic-offline-replica-rebuilding.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Automatic Offline Replica Rebuilding
 weight: 10
+aliases:
+- /spdk/features/automatic-offline-replica-rebuilding.md
```
innobead, line 5: Before 1.6, there was no `features` folder, so should the alias be `/spdk/automatic-offline-replica-rebuilding.md`?

innobead (changes requested):

content/docs/1.6.0/v2-data-engine/features/automatic-offline-replica-rebuilding.md
jillian-maroket: I have updated the alias.

innobead (approved):

PR #857, opened 2024-01-18
Reorganize nodes-and-volumes and advanced-resources

derekbit:

Issue #

Make overcrowded layout navigation bar collapsible and clear.

Before
image

After
image


derekbit:
The PR also tries to reorganize the chapters and sections. cc @innobead 

image

derekbit:
> `Uninstall Longhorn` seems not working. It should be a single page instead of a folder. Need to review other parts as well.

Weird. Checking. 

> `Uninstall Longhorn` seems not working. It should be a single page instead of a folder. Need to review other parts as well.

Fixed.
`Uninstall Longhorn` should be single page, but we put it in a folder before for making the `Uninstall Longhorn` page in the last one. You can see it will be 
image

innobead:
Like the change!

innobead (changes requested):
Can we display the sub-list after clicking the section title instead of only expanding it when clicking the + icon?

layouts/partials/docs/nav.html
```diff
@@ -66,22 +128,24 @@
...
-      {{ with .RegularPages }}
-      <ul class="nav-sub-section-pages">
-        {{ range . }}
-        {{ $isHere := eq $here .RelPermalink }}
-        <li class="nav-sub-section-pages-link">
-          <a{{ if $isHere }} class="is-active"{{ end }} href="{{ .RelPermalink }}">
-          {{ .Title }}
-          </a>
-        </li>
+      <details class="nav-section-details"{{ if $isHere }} close{{ end }} {{ if eq (len .RegularPages) 0 }}no-expand{{ end }}>
```
innobead, line 131: Suggest opening the detailed pages if the page is the current page, so that will be apparent for users to check them when navigating to the section page w/o another extra click.

`close` and `no-expand` seems customized attributes, so I am not sure if there would be some side-effects when displaying the page in different browsers? from my experience, probably it should be fine, but also can consider using `data-*` for customized attributes, [ref](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).

innobead (changes requested):
In general, LGTM. Just a few feedback.

derekbit:
I think we can replace the current theme with a better theme later (I need to study it a bit more).
I will make the PR focusing on the reorganization of chapters and sections.
WDYT? @innobead

innobead:
> I think we can replace the current theme with a better theme later (I need to study it a bit more). I will make the PR focusing on the reorganization of chapters and sections. WDYT? @innobead

Sounds good to me, but this was a good try.

innobead (approved):

PR #863, opened 2024-01-24
doc(ha): add doc for remount when volume read only

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/6386

Add doc for remount when volume becomes read only

### DIFF
--- a/content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
+++ b/content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
@@ -1,0 +1,14 @@
+---
+title: Recover Volume When It Becomes Read Only
+weight: 1
+---
+
+The state of a volume can change to read-only when IO errors occur. IO errors can be caused by a variety of issues, including the following:
+- Network disconnection: Interrupted connection between the engine and replicas
+- High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
+
+Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
+
+To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
+
+Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
--- a/content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
+++ b/content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
@@ -1,0 +1,14 @@
+---
+title: Recover Volume When It Becomes Read Only
+weight: 1
+---
+
+The state of a volume can change to read-only when IO errors occur. IO errors can be caused by a variety of issues, including the following:
+- Network disconnection: Interrupted connection between the engine and replicas
+- High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
+
+Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
+
+To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
+
+Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
--- a/content/docs/1.5.4/high-availability/recover-volume-when-read-only.md
+++ b/content/docs/1.5.4/high-availability/recover-volume-when-read-only.md
@@ -1,0 +1,14 @@
+---
+title: Recover Volume When It Becomes Read Only
+weight: 1
+---
+
+The state of a volume can change to read-only when IO errors occur. IO errors can be caused by a variety of issues, including the following:
+- Network disconnection: Interrupted connection between the engine and replicas
+- High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
+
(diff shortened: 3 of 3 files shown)


content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,13 @@
+---
+  title: Recover Volume When It Becomes Read Only
+  weight: 1
+---
+
+When an unexpected detachment happens, which can happen during a [Kubernetes upgrade](https://github.com/longhorn/longhorn/issues/703), a [Docker reboot](https://github.com/longhorn/longhorn/issues/686), or a network disconnection, the volume could become read only and leads to IO error.
```
derekbit, line 6:
```suggestion
When an unexpected detachment happens
```

A volume is converted to read-only state after IO requests encountered IO errors rather than detachment.
IO errors can be caused by a network disconnection, high disk latency, abnormal node shutdown and so on.

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,13 @@
+---
+  title: Recover Volume When It Becomes Read Only
+  weight: 1
+---
+
+A volume is converted to read-only state after IO requests encountered IO errors which can be caused by a network disconnection, high disk latency, abnormal node shutdown and so on.
```
jillian-maroket, line 6:
```suggestion
The state of a volume can change to read-only when IO errors occur. IO errors can be caused by network disconnection, high disk latency, abnormal node shutdown, and other issues.
```

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,13 @@
...
+Longhorn periodically detects the status of the volume and automatically deletes the workload pod if the pod is managed by a controller (e.g. deployment, statefulset, daemonset, etc...) when the volume becomes read only.
+By deleting the pod, its controller restarts the pod and Kubernetes handles volume reattachment and remount.
```
jillian-maroket, line 8:
```suggestion
Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
```

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,13 @@
...
+
+If you don't want Longhorn to automatically delete the workload pod, you can set it in [the setting `Automatically Delete Workload Pod when The Volume Is Detached Unexpectedly`](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly) in Longhorn UI.
```
jillian-maroket, line 10:
```suggestion
To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
```

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,13 @@
...
+
+For the pods that don't have a controller, Longhorn doesn't delete them because if Longhorn does, no one will restart them.
+To recover read only volumes, you would have to manually delete and recreate the pods that don't have a controller.
```
jillian-maroket, line 13:
```suggestion
Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
```

jillian-maroket (commented):
@ChanYiLin Let me know if you have concerns about the changes.

ChanYiLin:
Hi @jillian-maroket , thanks for the advice!
I have update the doc.

content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,11 @@
+---
+  title: Recover Volume When It Becomes Read Only
+  weight: 1
```
innobead, line 3: Remove needless indent for each variable definition.

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
```diff
@@ -0,0 +1,11 @@
+---
+  title: Recover Volume When It Becomes Read Only
+  weight: 1
+---
+
+The state of a volume can change to read-only when IO errors occur. IO errors can be caused by network disconnection, high disk latency, abnormal node shutdown, and other issues.
```
innobead, line 6: Suggest briefly describing the cases. 

network disconnection: between engine and replicas
high-disk latency: latency between replica and the corresponding disk
abnormal node shutdown: node down where the engine is running.

innobead (changes requested):
In general, LGTM. Just a few feedback.

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
jillian-maroket, line 6:
```suggestion
The state of a volume can change to read-only when IO errors occur. IO errors can be caused by a variety of issues, including the following:

- Network disconnection: Interrupted connection between the engine and replicas
- High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
- Abnormal node shutdown: Failure of a node that is running the engine
```

jillian-maroket (commented):
@ChanYiLin I rewrote the first paragraph and added a list.

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
innobead: @derekbit are there any missing or incorrect changes about the causes of IO errors? :point_up:

content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
ChanYiLin: Fixed.

content/docs/1.6.0/high-availability/recover-volume-when-read-only.md
ChanYiLin: Fixed.

derekbit: > Abnormal node shutdown: Failure of a node that is running the engine

When an engine is somehow shutdown, application and engine are both gone, so the IO error shouldn't happen.

> abnormal node shutdown

`abnormal node shutdown` in my previous comment might be not necessary.

WDYT? @ChanYiLin @innobead

ChanYiLin: Yeah agree, and I think here we just want to provide some hints about why it happens
But IO errors may be caused by many factors, I don't think we can and need to  list all of them.

ChanYiLin: Fixed.

jillian-maroket (approved):

innobead (approved):

PR #866, opened 2024-02-03
Correct setting name for disabling pre-upgrade checker

dolaru:
The value seems to be outdated. The pre-upgrade checker job is actually `preUpgradeChecker.jobEnabled: false` when using the Helm chart for `v1.6.0`.


innobead:
@mantissahz

mantissahz (approved):
LGTM and verified.

jillian-maroket (approved):
LGTM. Thanks!

innobead (approved):

PR #869, opened 2024-02-07
doc: update the behavior of remount read only volume

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/7843

change the behavior of the auto remount read only volume documents

### DIFF
--- a/content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
+++ b/content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
@@ -7,8 +7,4 @@ The state of a volume can change to read-only when IO errors occur. IO errors ca
 - Network disconnection: Interrupted connection between the engine and replicas
 - High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
 
-Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
-
-To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
-
-Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
+Longhorn periodically checks the volume state. When the filesystem of the volume becomes Read-only, Longhorn automatically attempts to remount the global mount point on the host to change the state back to read-write. On successful remount, the workload pods will continue working without service disruption.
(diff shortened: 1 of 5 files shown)


jillian-maroket (commented):
@ChanYiLin I split the original sentence into two so that the meaning is clearer. Please check if my understanding is correct.

content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
```diff
@@ -7,8 +7,4 @@ The state of a volume can change to read-only when IO errors occur. IO errors ca
...
 - High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
 
-Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
-
-To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
-
-Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
+Longhorn periodically checks the volume state and automatically remount the global mount point on the host to read write again when the state changes to read-only.
```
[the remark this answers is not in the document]
ChanYiLin: Yes, the context is correct.
Fixed, thanks for the advice!

content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
```diff
@@ -7,8 +7,4 @@ The state of a volume can change to read-only when IO errors occur. IO errors ca
...
 - High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk
 
-Longhorn periodically checks the volume state and automatically deletes the workload pod if the pod is managed by a controller (for example, Deployment, StatefulSet, or DaemonSet) when the state changes to read-only. Once the workload pod is deleted, the controller restarts the pod and Kubernetes handles volume reattachment and remounting.
-
-To prevent Longhorn from automatically deleting the workload pod, you can disable the setting on the Longhorn UI. For more information, see [Automatically Delete Workload Pod When the Volume Is Detached Unexpectedly](../../references/settings#automatically-delete-workload-pod-when-the-volume-is-detached-unexpectedly).
-
-Longhorn does not delete pods without a controller because such pods cannot be restarted after deletion. To recover read-only volumes, you must manually delete and restart the pods that are not managed by a controller.
+Longhorn periodically checks the volume state. When the state changes to read-only, Longhorn automatically remounts the global mount point on the host to change the state back to read-write.
```
innobead, line 10: Suggest 

> Longhorn periodically checks the volume state. When the filesystem of the volume becomes Read-only, Longhorn automatically attempts to remount the global mount point on the host to change the state back to read-write. On successful remount, the workload pods will continue working without service disruption.

BTW, do we have any place to explain when remount will be requested and the workload will be deleted accordingly? (like auto salvage, etc.) If no doc, we should take this chance to improve this part.

innobead (changes requested):

content/docs/1.4.5/high-availability/recover-volume-when-read-only.md
ChanYiLin: Do you mean this one
https://longhorn.io/docs/1.6.0/high-availability/recover-volume/

ChanYiLin: Fixed, thanks!

innobead (approved):

PR #871, opened 2024-02-08
Add Snapshot Maximum Count to the Settings Reference page

yardenshoham:
It was missing

Fixes https://github.com/longhorn/longhorn/issues/7894


PhanLe1010 (approved):
LGTM

PR #879, opened 2024-03-02
kb: list items using table instead of card

derekbit:

Issue #

We have more KB articles over time, to improve readability and searchability, use a table instead of cards.

Before
image

After 
image


innobead (approved):

innobead:
@derekbit Good improvement!

PR #890, opened 2024-03-27
release 1.6.1

innobead:
ref: https://github.com/longhorn/longhorn/issues/8209

- **docs: remove 1.4.5-dev doc**
- **docs: move older 1.4 docs to archives**
- **docs: add 1.6.2-dev doc**
- **docs: remove older 1.1 and 1.2 docs from archives**
- **docs: move all 1.4 to archives**


innobead:
In this PR, the primary change is to make the 1.6.1 doc the latest release doc. I also moved 1.4 docs to the archives and removed some unsupported documents.

PR #892, opened 2024-04-09
Update install-with-argocd.md

lexfrei:
Update outdated example


jillian-maroket (approved):
@lexfrei LGTM. Thank you for catching the error and for contributing to the doc.

derekbit:
Thanks @lexfrei.
It is fixed in https://github.com/longhorn/website/pull/894.

PR #900, opened 2024-04-28
v2 volume: support filesystem trim

derekbit:

Issue Longhorn/longhorn#7534


innobead (changes requested):
Can link to the subpage.

innobead (approved):

PR #915, opened 2024-05-27
doc(install-argocd): updated to the right helm values

cchriso:
the helm values changed and are not longer correct, so I corrected them according to the official helm chart https://github.com/longhorn/charts/blob/v1.6.x/charts/longhorn/values.yaml#L165

Issue #


innobead (approved):

innobead:
@cchriso thanks for the contribution!

@mantissahz let's ensure the doc up to date.

cc @derekbit

PR #916, opened 2024-05-28
doc(1.7.0/1.6.3/1.5.6): add support bundle node collection timeout setting

c3y1huang:

Issues:
- longhorn/longhorn#8623
- longhorn/longhorn#8624
- longhorn/longhorn#8625

Add setting reference for new `support-bundle-node-collection-timeout` setting.

`None`

`None`

### DIFF
--- a/content/docs/1.5.6/references/settings.md
+++ b/content/docs/1.5.6/references/settings.md
@@ -32,6 +32,7 @@ weight: 1
   - [Engine to Replica Timeout](#engine-to-replica-timeout)
   - [Support Bundle Manager Image](#support-bundle-manager-image)
   - [Support Bundle Failed History Limit](#support-bundle-failed-history-limit)
+  - [Support Bundle Node Collection Timeout](#support-bundle-node-collection-timeout)
   - [Fast Replica Rebuild Enabled](#fast-replica-rebuild-enabled)
   - [Timeout of HTTP Client to Replica File Sync Server](#timeout-of-http-client-to-replica-file-sync-server)
 - [V2 Data Engine (Preview Feature)](#v2-data-engine-preview-feature)
@@ -317,6 +318,7 @@ This information will help us gain insights how Longhorn is being used, which wi
     - Storage Over Provisioning Percentage
     - Storage Reserved Percentage For Default Disk
     - Support Bundle Failed History Limit
+    - Support Bundle Node Collection Timeout
     - System Managed Pods Image Pull Policy
 
 > The `Upgrade Checker` needs to be enabled to periodically send the collected data.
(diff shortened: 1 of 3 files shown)


content/docs/1.5.6/references/settings.md
```diff
@@ -399,6 +401,14 @@ The retained failed support bundle is for analysis purposes and needs to clean u
 
 Longhorn blocks support bundle creation when reaching the upper bound of the limitation. You can set this value to **0** to have Longhorn automatically purge all failed support bundles.
 
+#### Support Bundle Node Collection Timeout
+
+> Default: `30`
+
+The value in minutes specifies the timeout of node bundle collection for support bundle generation.
+
+When the timeout is reached, the support bundle generation will proceed without requiring the collection of node bundles.
```
jillian-maroket, line 410:
```suggestion
Number of minutes Longhorn allows for collection of node data for the support bundle.

If the collection process is not completed within the allotted time, Longhorn continues generating the support bundle without the uncollected node data.
```

jillian-maroket: Questions:
- What type of data is collected for the "node bundle"? I'm asking because [Harvester](https://github.com/harvester/docs/pull/569/files) is more specific about the collection scope.
- Are there minimum and maximum values for this setting? What happens if the user specifies a value less than 30?

jillian-maroket (commented):
@c3y1huang I listed a few questions that might help us improve the description.

content/docs/1.5.6/references/settings.md
c3y1huang: > * What type of data is collected for the "node bundle"? I'm asking because [Harvester](https://github.com/harvester/docs/pull/569/files) is more specific about the collection scope.

The node agent currently collects the following, and there are also plans to include additional information and logs.
- host system info
- dmesg log
- host system log
- kubelet log

> * Are there minimum and maximum values for this setting? What happens if the user specifies a value less than 30?

The minimum value is 0, with no maximum limit. If the user specifies a value less than 30, the timeout will be set to that value in minutes.

jillian-maroket: Thank you. Can we be more specific (say "node configuration and logs") in the description? This is to emphasize that we are not collecting user data.

*Number of minutes Longhorn allows for collection of node configuration and logs for the support bundle.*

*If the collection process is not completed within the allotted time, Longhorn continues generating the support bundle without the uncollected node data.*

jillian-maroket (approved):
Thanks for the update, will approve now. You can also consider changing *node information and node logs* to *node configuration and logs* if that wording is correct. It's a shorter and more specific.

content/docs/1.5.6/references/settings.md
jillian-maroket: Thank you for the clarification.

c3y1huang:
> Thanks for the update, will approve now. You can also consider changing _node information and node logs_ to _node configuration and logs_ if that wording is correct. It's a shorter and more specific.

I don't think _node configuration_ is the right word for what the node agent gathers for Longhorn, which is different from Harvester. Let's leave it as _node information_ and _node log_ for now.

PR #917, opened 2024-05-28
Document minimum XFS volume size backport

ejweber:

longhorn/longhorn#8488

I was previously not planning on backporting the changes for longhorn/longhorn#8488. Now that we are backporting the changes, they need to be documented in the appropriate releases.


jillian-maroket (approved):
LGTM. Thanks!

PR #925, opened 2024-06-24
Make long gRPC timeout configurable

PhanLe1010:
longhorn/longhorn#2765


jillian-maroket (approved):

content/docs/1.7.0/references/settings.md
```diff
@@ -456,6 +457,12 @@ The setting enables fast replica rebuilding feature. It relies on the checksums
 
 The value in seconds specifies the timeout of the HTTP client to the replica's file sync server used for replica rebuilding, volume cloning, snapshot cloning, etc.
 
+#### Long gRPC Timeout
```
innobead, line 460: nit: Just noticed that the setting name is quite common, so it's not easy to determine its meaning directly. Perhaps it can be renamed to `Replica Sync Timeout` or something similar. What do you think?

innobead: Discussed with Phan. It seems a consensus made, so don't need to change here.

innobead (approved):

PR #934, opened 2024-07-01
Fix install with argocd targetRevision not match each version

tico88612:

Issue #

Original is static content, this PR fixed to `{{< current-version >}}`


jillian-maroket (approved):
Thank you for contributing!

PR #936, opened 2024-07-03
doc(v2 data engine): remove offline rebuilding

derekbit:

Issue Longhorn/longhorn#8442

offline rebuilding is removed and replaced with online rebuilding.


DamiaSan (approved):
Just one question: why are the settings of the `Danger zone` risky?

derekbit:
> Just one question: why are the settings of the `Danger zone` risky?

No harmful.
The change of danger zone layout is automatically updated by IDE and not related to the PR.

PR #952, opened 2024-07-25
update install doc to enable and start iscsi

abonillabeeche:

Updates the doc as the SUSE/OpenSUSE section does not cover enabling and starting the iscsid service

`longhorn-test-01:~ # systemctl enable iscsid
Created symlink /etc/systemd/system/multi-user.target.wants/iscsid.service → /usr/lib/systemd/system/iscsid.service.`


derekbit (approved):
Good catch! Thank you.

PR #954, opened 2024-07-26
doc(v1.7.0): update "Important Notes"

derekbit:

Issue Longhorn/longhorn#8813


innobead (approved):

innobead (commented):
Let's make sure environment setup and check change accordingly, ex: cli.

BTW, should we update the environment setup script to use longhornctl? Can still use that script but call the cli directly.

derekbit:
> Let's make sure environment setup and check change accordingly, ex: cli.

Yes, we have a ticket for updating the document
https://github.com/longhorn/longhorn/issues/9075

cc @mantissahz 

> BTW, should we update the environment setup script to use longhornctl? Can still use that script but call the cli directly.

Can we deprecate the environment setup script and start using `longhornctl`?
There seems to be no benefit in maintaining the script."

PR #956, opened 2024-07-29
docs(v1.7.0): Add note about possible RWX fast-failover deadlock situ…

james-munson:
…ation.

Warning mentioned in https://github.com/longhorn/longhorn/issues/9093#issuecomment-2256250764

### DIFF
--- a/content/docs/1.7.0/high-availability/rwx-volume-fast-failover.md
+++ b/content/docs/1.7.0/high-availability/rwx-volume-fast-failover.md
@@ -14,3 +14,9 @@ Along with adding the monitoring and fast reaction, the feature also changes the
 If the setting is changed back to "false", the lease check is disabled and pod relocation will use regular Kubernetes rules for node failure, even on existing volumes.  When the server pod is next restarted, it will revert to the normal grace period configuration.
 
 For more information, see https://github.com/longhorn/longhorn/issues/6205.
+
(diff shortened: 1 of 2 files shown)


derekbit (commented):
LGTM. Can you add this in the important node as well? Thank you.

https://longhorn.io/docs/1.7.0/deploy/important-notes/#rwx-volumes-fast-failover

content/docs/1.7.0/high-availability/rwx-volume-fast-failover.md
```diff
@@ -14,3 +14,9 @@ Along with adding the monitoring and fast reaction, the feature also changes the
 If the setting is changed back to "false", the lease check is disabled and pod relocation will use regular Kubernetes rules for node failure, even on existing volumes.  When the server pod is next restarted, it will revert to the normal grace period configuration.
 
 For more information, see https://github.com/longhorn/longhorn/issues/6205.
+
+> **Note:**  In rare circumstances, it is possible for the failover to become deadlocked, if the NFS server pod creation is blocked by a recovery action that is itself blocked by the failover-in-process state.  If that is the case, and failover takes more than a minute or two, the workaround is simply to delete the associated lease object.  That will clear the state, and a new lease will be created along with the replacement server pod.  For example, if the stuck volume is named `pvc-2ce4e82e-7ccc-46c0-90a8-a141501fbf93` and the feature is enabled, there will be a lease with the same name.  Simply do
```
jhkrug, line 18:
```suggestion
> **Note:**  In rare circumstances, it is possible for the failover to become deadlocked. This happens if the NFS server pod creation is blocked by a recovery action that is itself blocked by the failover-in-process state.  If that is the case, and failover takes more than a minute or two, the workaround is to delete the associated lease object.  That clears the state, and a new lease will be created along with the replacement server pod.  For example, if the stuck volume is named `pvc-2ce4e82e-7ccc-46c0-90a8-a141501fbf93` and the feature is enabled, there will be a lease with the same name.  To delete the associated lease object:
```

jhkrug (changes requested):
A couple of minor suggestions

PR #957, opened 2024-07-29
docs:  Update forcible CRD deletion script in uninstall page.

james-munson:

Issue https://github.com/longhorn/longhorn/issues/9116, `[DOC] Last-ditch uninstall cleanup instructions are out of date.`

### DIFF
--- a/content/docs/1.7.0/deploy/uninstall/_index.md
+++ b/content/docs/1.7.0/deploy/uninstall/_index.md
@@ -148,8 +148,8 @@ If your CRD instances or the CRDs themselves can't be deleted for whatever reaso
 
 ```shell
 # Delete CRD finalizers, instances and definitions
-for crd in $(kubectl get crd -o jsonpath={.items[*].metadata.name} | tr ' ' '\n' | grep longhorn.rancher.io); do
-  kubectl -n ${NAMESPACE} get $crd -o yaml | sed "s/\- longhorn.rancher.io//g" | kubectl apply -f -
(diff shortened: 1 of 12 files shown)


content/docs/1.7.0/deploy/uninstall/_index.md
```diff
@@ -142,18 +141,5 @@ you can cancel the uninstallation as the following.
...
-
-```shell
-# Delete CRD finalizers, instances and definitions
-for crd in $(kubectl get crd -o jsonpath={.items[*].metadata.name} | tr ' ' '\n' | grep longhorn.rancher.io); do
-  kubectl -n ${NAMESPACE} get $crd -o yaml | sed "s/\- longhorn.rancher.io//g" | kubectl apply -f -
-  kubectl -n ${NAMESPACE} delete $crd --all
-  kubectl delete crd/$crd
-done
-```
-
```
derekbit, line 157: Shall we delete the commands? Could we replace `longhorn.rancher.io` with `longhorn.io`?

jhkrug: I don't have enough technical context to be sure yet, but if the commands are no longer needed then deletion is good.

innobead: I think we should keep this since it is related to a case where the CRs somehow can't be deleted.

derekbit (dismissed):

derekbit:
Can you update the old versions as well?

jhkrug (approved):
lgtm

innobead (approved):

PR #963, opened 2024-08-08
doc(os): bump OS versions supported by Longhorn

derekbit:

Issue #

### DIFF
--- a/content/docs/1.5.0/best-practices.md
+++ b/content/docs/1.5.0/best-practices.md
@@ -49,9 +49,9 @@ The following Linux OS distributions and versions have been verified during the 
 | No. | OS           | Versions
 |-----|--------------| --------
 | 1.  | Ubuntu       | 22.04
-| 2.  | SLES         | 15 SP4
-| 3.  | SLE Micro    | 5.3
-| 4.  | RHEL         | 9.1
+| 2.  | SUSE Linux Enterprise Server         | 15 SP4
+| 3.  | SUSE Linux Enterprise Micro    | 5.3
+| 4.  | Red Hat Enterprise Linux         | 9.1
 | 5.  | Oracle Linux | 9.1
 | 6.  | Rocky Linux  | 9.2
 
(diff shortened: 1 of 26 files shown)


jillian-maroket (commented):
Nit: Use the full name of the distributions.

jillian-maroket (approved):

jillian-maroket (dismissed):

content/docs/1.6.0/advanced-resources/os-distro-specific/talos-linux-support.md
```diff
@@ -60,6 +60,41 @@ For detailed instructions, see the Talos documentation on [Editing Machine Confi
...
+When [upgrading a Talos Linux node](https://www.talos.dev/v1.7/talos-guides/upgrading-talos/#talosctl-upgrade), always include the `--preserve` option in the command. This option explicitly tells Talos to keep ephemeral data intact.
+
+Example:
+
+```
+talosctl upgrade --nodes 10.20.30.40 --image ghcr.io/siderolabs/installer:v1.7.6 --preserve
+```
+
+> **Caution:**
+> If you do not include the `--preserve` option, Talos wipes `/var/lib/longhorn`, destroying all replicas stored on that node.
```
innobead, line 74: The other option is to suggest that users set up a dedicated partition of a disk as a Longhorn disk. This will be soon used in v2 adoption.

https://www.talos.dev/v1.7/reference/configuration/v1alpha1/config/#Config.machine.disks.

@c3y1huang can you improve the doc later?

innobead (changes requested):
In general, LGTM. 

Because v1.5 and v1.6 are still supported, we need to update SLES and SLE Micro versions to the latest released version doc of each as well. 

Thus, please help update v1.5.5, v1.5.6, v1.6.2, v1.6.3 docs accordingly. 

cc @longhorn/qa

yangchiu (approved):
LGTM

content/docs/1.6.0/advanced-resources/os-distro-specific/talos-linux-support.md
derekbit: Remove the commits from this PR because they are not in the scope of the PR.

yangchiu (approved):

PR #964, opened 2024-08-08
doc(uninstall): Add tip about working around missing webhook service.

james-munson:

https://github.com/longhorn/longhorn/issues/9116

This is an addendum to the section on forcible cleanup.  The user may hit Longhorn webhooks that are still registered although the services have been uninstalled.  Here's how to fix that.


james-munson:
After review and any mods, I will duplicate this to all current release docs.

jillian-maroket (commented):
Let me know if you have concerns about the changes.

jillian-maroket (approved):

innobead (approved):

PR #971, opened 2024-08-19
doc: provides descriptions of the volume, engine, and replica condition.

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/8508

Add descriptions of the volume, engine, and replica condition.


content/docs/1.5.4/high-availability/recover-volume.md
```diff
@@ -27,7 +27,7 @@ The state of a volume can change to read-only when IO errors occur. IO errors ca
 - Network disconnection: Interrupted connection between the engine and replicas.
 - High disk latency: Significant delay in the transfer of data between a replica and the corresponding disk.
 
-Longhorn checks the state of the volume's global mount point every 10 seconds. When the volume's filesystem changes to read-only, Longhorn updates the condition to the volume's data engine. Longhorn then automatically attempts to remount the global mount point on the host to change the state back to read-write. Upon successful remounting, the workload pods continue functioning without disruption.
+Longhorn checks the state of the volume's global mount point every 10 seconds. When the volume's filesystem changes to read-only, Longhorn updates the condition to the volume's data engine. Longhorn then automatically attempts to remount the global mount point on the host to change the state back to read-write. Upon successful remounting, the workload pods continue functioning without disruption. However, if the mount point becomes write-protected and Longhorn fails to remount the mount point, you may still need to manually recreate the workload to force it reattach and remount the volume.
```
jillian-maroket, line 30:
```suggestion
Longhorn checks the state of the volume's global mount point every 10 seconds. When the volume's filesystem changes to read-only, Longhorn updates the condition to the volume's data engine. Longhorn then automatically attempts to remount the global mount point on the host to change the state back to read-write. Upon successful remounting, the workload pods continue functioning without disruption. Longhorn also attempts to remount when the volume's filesystem changes to write-protected. If the remounting attempt fails, you may need to manually recreate the workload to force it reattach and remount the volume. 
```

jillian-maroket: The sentence is too long so I separated the system action and user action.

jillian-maroket (commented):
Initial review done. Only the "Volume Conditions" section has an introductory sentence. Can we provide more info about "Engine Conditions" and "Replica Conditions"?

content/docs/1.5.4/high-availability/recover-volume.md
ChanYiLin: Hi @jillian-maroket 
I feel like the meaning has changed. 
Longhorn remount the mount point if it is now read-only.
However, the mount point can be read-only and also write protected. In this case, Longhorn won't be able to remount it to read-write.

That is, Longhorn only remount the mount point if it is read-only. The "write protected" is the case that Longhorn won't be remount the mount point.

jillian-maroket: @ChanYiLin Thank you for clarifying. Let's use your original wording. I can't think of a way to separate the many concepts without altering the meaning.

content/docs/1.6.0/nodes-and-volumes/volumes/volume-conditions.md
```diff
@@ -0,0 +1,27 @@
...
+- `Scheduled`: All replicas were scheduled successfully. 
+    If Longhorn was unable to schedule any of the replicas, the condition is set to `false` and error messages are displayed. The condition is set to `true` when the setting [Allow Volume Creation With Degraded Availability](../../../references/settings#allow-volume-creation-with-degraded-availability) is enabled and at least one replica is scheduled.
+- `TooManySnapshots`: This specific volume has more than 100 snapshots.
+    Longhorn allows you to create a maximum of 250 snapshots for each volume. For more information about configuring the maximum snapshot count, see [Snapshot Space Management](../../../snapshots-and-backups/snapshot-space-management).
+- `Restore`: Longhorn is restoring the volume from a backup.
+- `WaitForBackingImage`: The replicas have not started because the backing images must first be synced with their disks.
+
+## Engine Conditions
+
+Engine conditions describe the current status of a engine and potential issues that may occur.
```
jillian-maroket, line 18:
```suggestion
Engine conditions describe the current status of an engine instance and potential issues that may occur.
```

jillian-maroket: Are you are referring to the engine instance on the node the volume is attached to? If not, you can ignore this suggestion.

jillian-maroket (commented):
Please check my comments.

content/docs/1.6.0/nodes-and-volumes/volumes/volume-conditions.md
ChanYiLin: not the instance but the resource in kubernetes
`$kubectl get lhe -n longhorn-system`  (get longhorn engine CR)

ChanYiLin: fixed, thanks

jillian-maroket (approved):

PR #975, opened 2024-08-23
Remove downgrade examples from upgrade path table

jillian-maroket:
Issue #9291

This PR removes downgrade examples from the table that explains the supported upgrade paths. Longhorn does not support downgrades at all so those examples do not provide any value to users and may even cause confusion.


derekbit (approved):
LGTM

PR #978, opened 2024-08-26
docs: describe new last replica timeout behavior

ejweber:
Longhorn 8711

longhorn/longhorn#8711

We need to clarify how the behavior of the `engine-replica-timeout` setting changes in v1.7.1 and v1.8.0.

Per https://github.com/longhorn/longhorn/wiki/Longhorn-Setting-Update, the actual setting definition stored in longhorn-manager should continue to be a one sentence description, so I think we only need to make the change here in the website.

I changed the heading from `Engine to Replica Timeout` to `Engine Replica Timeout` because the actual setting name is `engine-replica-timeout` and I think the "to" might be confusing to users looking for the setting to modify in their environments.

### DIFF
--- a/content/docs/1.7.1/references/settings.md
+++ b/content/docs/1.7.1/references/settings.md
@@ -32,7 +32,7 @@ weight: 1
   - [Backing Image Cleanup Wait Interval](#backing-image-cleanup-wait-interval)
   - [Backing Image Recovery Wait Interval](#backing-image-recovery-wait-interval)
   - [Default Min Number Of Backing Image Copies](#default-min-number-of-backing-image-copies)
-  - [Engine to Replica Timeout](#engine-to-replica-timeout)
+  - [Engine Replica Timeout](#engine-replica-timeout)
   - [Support Bundle Manager Image](#support-bundle-manager-image)
   - [Support Bundle Failed History Limit](#support-bundle-failed-history-limit)
   - [Support Bundle Node Collection Timeout](#support-bundle-node-collection-timeout)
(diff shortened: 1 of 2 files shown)


content/docs/1.7.1/references/settings.md
```diff
@@ -417,10 +417,21 @@ The interval in seconds determines how long Longhorn will wait before re-downloa
...
 The default minimum number of backing image copies Longhorn maintains.
 
-#### Engine to Replica Timeout
+#### Engine Replica Timeout
+
 > Default: `8`
 
-The value in seconds specifies the timeout of the engine to the replica(s), and the value should be between 8 to 30 seconds.
+The time in seconds a v1 engine will wait for a response from a replica before marking it as failed. Values between 8
+and 30 are allowed. Only in effect while there are I/O requests outstanding.
```
jhkrug, line 425:
```suggestion
and 30 are allowed. The engine replica timeout is only in effect while there are I/O requests outstanding.
```

content/docs/1.7.1/references/settings.md
```diff
@@ -417,10 +417,21 @@ The interval in seconds determines how long Longhorn will wait before re-downloa
...
+
+This timeout only applies as-configured to additional/extra replicas. A v1 engine will not mark the final replica for a
```
jhkrug, line 427: Only one of 'additional' or 'extra' is needed.
```suggestion
This timeout only applies to additional replicas. A v1 engine will not mark the final replica for a
```

jhkrug (changes requested):
If the suggestions are acceptable they are needed in 1.8.0 as well. Thanks.

ejweber:
I made the requested changes in an amend.

derekbit (approved):
LGTM

derekbit (approved):

derekbit:
@ejweber Can you add it into the important notes? Thank you.

derekbit (approved):

jhkrug (changes requested):
A suggestion.

ejweber:
Sorry @jhkrug. I was going back-and-forth about removing "as-configured", but in the end I decided to just follow your suggestion without comment. (I would not have resolved the conversation without discussion if I wasn't going to make the suggested changes directly.) It seems I messed up and pushed a version of my amend that didn't have your suggestion.

Looks like I need to update the important notes anyway, so we can catch it there.

content/docs/1.7.1/references/settings.md
```diff
@@ -417,10 +417,21 @@ The interval in seconds determines how long Longhorn will wait before re-downloa
...
-#### Engine to Replica Timeout
+#### Engine Replica Timeout
+
 > Default: `8`
 
-The value in seconds specifies the timeout of the engine to the replica(s), and the value should be between 8 to 30 seconds.
+The time in seconds a v1 engine will wait for a response from a replica before marking it as failed. Values between 8
+and 30 are allowed. The engine replica timeout is only in effect while there are I/O requests outstanding.
+
+This timeout only applies as-configured to additional replicas. A v1 engine will not mark the final replica for a
```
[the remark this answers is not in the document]
ejweber: Sorry @jhkrug. I was going back-and-forth about removing "as-configured", but in the end I decided to just follow your suggestion without comment. (I would not have resolved the conversation without discussion if I wasn't going to make the suggested changes directly.) It seems I messed up and pushed a version of my amend that didn't include the suggestion after all.

Looks like I need to update the important notes anyway, so we can catch it there.

https://github.com/longhorn/website/pull/980

PR #987, opened 2024-09-09
doc(setting): modify the setting description

mantissahz:

Issue # longhorn/longhorn#9391

Modify the setting `Default Longhorn Static Storage Class Name` description.

### DIFF
--- a/content/docs/1.7.2/references/settings.md
+++ b/content/docs/1.7.2/references/settings.md
@@ -232,7 +232,7 @@ Every Longhorn release will ship with a new Longhorn engine image. If the curren
 
 > Default: `longhorn-static`
 
-The `storageClassName` is for persistent volumes (PVs) and persistent volume claims (PVCs) when creating PV/PVC for an existing Longhorn volume. Notice that it's unnecessary for users to create the related StorageClass object in Kubernetes since the StorageClass would only be used as matching labels for PVC bounding purpose. By default 'longhorn-static'.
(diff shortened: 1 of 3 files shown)


content/docs/1.7.2/references/settings.md
```diff
@@ -232,7 +232,7 @@ Every Longhorn release will ship with a new Longhorn engine image. If the curren
 
 > Default: `longhorn-static`
 
-The `storageClassName` is for persistent volumes (PVs) and persistent volume claims (PVCs) when creating PV/PVC for an existing Longhorn volume. Notice that it's unnecessary for users to create the related StorageClass object in Kubernetes since the StorageClass would only be used as matching labels for PVC bounding purpose. By default 'longhorn-static'.
+The `storageClassName` is for persistent volumes (PVs) and persistent volume claims (PVCs) when creating PV/PVC for an existing Longhorn volume. Notice that it's unnecessary for users to create the related StorageClass object in Kubernetes since the StorageClass would only be used as matching labels for PVC bounding purpose.  The "storageClassName" needs to be an existing StorageClass. Only the StorageClass named `longhorn-static` will be create if it does not exist. By default 'longhorn-static'.
```
jhkrug, line 235:
```suggestion
The `storageClassName` is for persistent volumes (PVs) and persistent volume claims (PVCs) when creating PV/PVC for an existing Longhorn volume. Notice that it's unnecessary for users to create the related StorageClass object in Kubernetes since the StorageClass would only be used as matching labels for PVC bounding purposes.  The "storageClassName" needs to be an existing StorageClass. Only the StorageClass named `longhorn-static` will be created if it does not exist. By default 'longhorn-static'.
```

jhkrug (changes requested):
A couple of typo corrections.

content/docs/1.7.2/references/settings.md
mantissahz: Modified. Thanks.

derekbit:
@mantissahz Need to update v1.8.0-dev and v1.6.3 as well. Thank you.

mantissahz:
>@mantissahz Need to update v1.8.0-dev and v1.6.3-dev as well. Thank you.

Added. Thanks

derekbit (approved):

PR #988, opened 2024-09-11
docs(rwx): Fix mistaken note about auto-deletion setting

james-munson:

https://github.com/longhorn/longhorn/issues/9415

Correct a statement about Longhorn restart of RWX workload pods when auto-delete-pod-when-volume-detached-unexpectedly is true.


jhkrug (approved):
ok

PR #1016, opened 2024-12-06
docs: correct module name from nvme-tcp to nvme_tcp

jangseon-ryu:

Issue #

The kernel module name 'nvme_tcp' was incorrectly documented as 'nvme-tcp'. 
This commit corrects the module name to ensure consistency with the actual kernel module naming convention.


derekbit (approved):
@jangseon-ryu Thanks for your contribuition!

PR #1017, opened 2024-12-06
doc(deploy): mount host's directory /boot

mantissahz:

Issue # longhorn/longhorn#9153

Update the document that we add a new directory.

### DIFF
--- a/content/docs/1.7.2/deploy/install/_index.md
+++ b/content/docs/1.7.2/deploy/install/_index.md
@@ -156,6 +156,7 @@ Longhorn components require root access with privileged permissions to achieve v
 Below are the directories Longhorn components requiring access with root and privileged permissions :
 
 - Longhorn Manager
+  - /boot: Get required modules' information from /boot/config-$(uname -r) on the host.
   - /dev: Block devices created by Longhorn are under the `/dev` path.
   - /proc: Find the recognized host process like container runtime, then use `nsenter` to access the mounts on the host to understand disks usage.
   - /var/lib/longhorn: The default path for storing volume data on a host.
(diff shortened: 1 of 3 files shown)


derekbit:
@jillian-maroket @jhkrug Please help review the PR. Thanks.

content/docs/1.7.2/deploy/install/_index.md
```diff
@@ -156,6 +156,7 @@ Longhorn components require root access with privileged permissions to achieve v
 Below are the directories Longhorn components requiring access with root and privileged permissions :
 
 - Longhorn Manager
+  - /boot: Get required modules' information from /boot/config-$(uname -r) on the host.
```
innobead, line 159: Please make sure the mount is read-only in the implementation.

It seems more host mounts are required for environment check purposes. We need to ensure the least privilege.

innobead: It seems we are adding similar capacities of CLI preflight to the longhorn-manager. Will this cause duplicated implementation? Why not just use CLI directly?

derekbit: CLI preflight check is usually executed before installing Longhorn system, and longhorn-manager node.status.condition is for runtime self-diagnosis. Although they check the same kernel modules and utilities, they serve different purposes.

derekbit: WDYT? @innobead @c3y1huang

innobead: This has been discussed with @derekbit . What I care about is only duplicated implementation, so if we can achieve that in the same shared code eventually, that will be good.

jillian-maroket (approved):

content/docs/1.7.2/deploy/install/_index.md
c3y1huang: https://github.com/longhorn/longhorn/issues/9933

c3y1huang (approved):

PR #1049, opened 2025-02-04
doc: add doc for accessing ui through rancher when network policy is enabled.

ChanYiLin:
There is something needs to be configured when accessing the Longhorn UI through Rancher when the network policy is enabled.


jhkrug (approved):
lgtm

derekbit (approved):
LGTM

PR #1053, opened 2025-02-07
docs: troubleshooting two active engines during volume live migration

PhanLe1010:
longhorn/longhorn#6642


innobead (approved):
Technically LGTM.

Need @longhorn/doc review.

Vicente-Cheng (approved):
lgtm, very clear for the workaround. Thanks!

derekbit (approved):

PR #1045, opened 2025-01-21
doc(1.9.0): recurring system backup

c3y1huang:
#### Which issue(s) this PR fixes:

Issue longhorn/longhorn#6534

#### What this PR does / why we need it:

- Add to important note.
- Add steps for creating recurring system backups.

#### Special notes for your reviewer:

`None`

#### Additional documentation or context

`None`

PR #940, opened 2024-07-08
doc(v1.7.0): add clone volume features from longhorn UI

a110605:
#### Which issue(s) this PR fixes:

Issue # https://github.com/longhorn/longhorn/issues/8741

#### What this PR does / why we need it:
Add block description how to clone volume from UI

#### Special notes for your reviewer:
See screen video in  https://github.com/longhorn/longhorn-ui/pull/760

PR #1054, opened 2025-02-10
doc(1.9.0): fix convention and formatting

c3y1huang:

Issue longhorn/longhorn#6534

https://github.com/longhorn/website/pull/1045#discussion_r1948675253

`None`

`None`

### DIFF
--- a/content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
+++ b/content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
@@ -77,7 +77,7 @@ The Longhorn system backup offers the following volume backup policies:
 
 #### Create a System Backup Using the Longhorn UI
 
-1. Go to the `System Backup` page in the `Setting` drop-down list.
+1. Go to the `System Backup` screen in the `Setting` drop-down list.
 1. Click `Create` under `System Backup`.
 1. Give a `Name` for the system backup.
 1. Select a `Volume Backup Policy` for the system backup.
@@ -106,7 +106,7 @@ The Longhorn system backup offers the following volume backup policies:
 
 #### Create a Recurring Backup Job Using the Longhorn UI
 
-1. Go to the `Recurring Job` page.
+1. Go to the `Recurring Job` screen.
 1. Click on `Create Recurring Job`.
 1. Configure the following settings:
    - **Name**: Specify a name for the recurring job.
@@ -145,7 +145,7 @@ You can delete the Longhorn system backup in the remote backup target using the 
 
 ### Delete a System Backup Using the Longhorn UI
 
-1. Go to the `System Backup` page in the `Setting` drop-down list.
+1. Go to the `System Backup` screen in the `Setting` drop-down list.
 1. Delete a single system backup in the `Operation` drop-down menu next to the system backup. Or delete in batch with the `Delete` button.
 
    > **Note:** Deleting the system backup will also make a deletion in the backup store.
--- a/content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
+++ b/content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
@@ -78,9 +78,13 @@ The Longhorn system backup offers the following volume backup policies:
 #### Create a System Backup Using the Longhorn UI
 
 1. Go to the `System Backup` screen in the `Setting` drop-down list.
+
 1. Click `Create` under `System Backup`.
+
 1. Give a `Name` for the system backup.
+
 1. Select a `Volume Backup Policy` for the system backup.
+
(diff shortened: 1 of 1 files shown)


content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
```diff
@@ -106,14 +111,17 @@ The Longhorn system backup offers the following volume backup policies:
 
 #### Create a Recurring Backup Job Using the Longhorn UI
 
-1. Go to the `Recurring Job` page.
```
c3y1huang, line 109: The term `screen` is unfamiliar to me when it comes to navigating the UI.

cc @jillian-maroket @derekbit @innobead

content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
```diff
@@ -106,14 +111,17 @@ The Longhorn system backup offers the following volume backup policies:
...
+1. Go to the `Recurring Job` screen.
```
c3y1huang, line 114: The term `screen` is unfamiliar to me when it comes to navigating the UI.

cc @jillian-maroket @derekbit @innobead

c3y1huang:
Converted to `draft`. Waiting for confirmation from @derekbit .

content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
```diff
@@ -77,10 +77,14 @@ The Longhorn system backup offers the following volume backup policies:
 
 #### Create a System Backup Using the Longhorn UI
 
-1. Go to the `System Backup` page in the `Setting` drop-down list.
+1. Go to the `System Backup` screen in the `Setting` drop-down list.
+
```
c3y1huang, line 81: I don't understand why we needed extra blank lines between the numbering points. I believe we usually avoid adding extra line breaks because we follow the Markdown [ordered lists](https://www.markdownguide.org/basic-syntax/#ordered-lists) format, which prevents the numbering from breaking. However, even with a line break, it doesn’t seem to affect the formatting on the website. So, is the suggested change for internal codebase management?

content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
c3y1huang: @jillian-maroket 
Could you clarify why `screen` is the right term for the UI context? When we guide the users to a specific section, we refer to Longhorn's webpage rather than the entire screen. That's where I am confused with the term.

https://github.com/longhorn/website/pull/940#issuecomment-2224752899
> Use page or webpage to describe one of a collection of web documents that make up a website. Use page to refer to the page the reader is on or to a specific page, such as the home page or start page.

> Use instead of screenful or full screen.

jillian-maroket: Longhorn's website, which has web pages, is different from its web-based management console (what the team calls the "Longhorn UI"), which has screens. Even the name "Longhorn UI" is not appropriate because Longhorn now has a CLI, which is a type of UI. Essentially, Longhorn has 2 UIs: the GUI called "Longhorn UI" and the CLI called "longhornctl". 

Perhaps you are unfamiliar with the term "screen" because you don't usually work with UX designers (who are different from UI engineers like Andy and Yiya). I worked with UX professionals before joining SUSE and we always used "screen". If you don't accept my explanations, maybe you'll believe Google?

https://m3.material.io/foundations/layout/understanding-layout/overview

@innobead @derekbit This is the last time that I will be justifying this term's usage to members of this team.

jillian-maroket (approved):

content/docs/1.9.0/advanced-resources/system-backup-restore/backup-longhorn-system.md
c3y1huang: @jillian-maroket and I have discussed this, and before the merge, I'd like to reach a consensus. I will bring this up during the Longhorn sync-up meeting.

For a web-based application like Longhorn UI, I believe `page` is the more appropriate term for navigation. In the context of web apps, `page` typically refers to different views or sections users navigate to via links or menus. On the other hand, `screen` is more commonly associated with mobile apps or devices with smaller, fixed interfaces.

Most users are familiar with the term `page`, and I don't think we should assume they have experience working with UX designers.

Personally, I prefer `page`, and I think it will be more intuitive for most users. Since users are interacting with a web browser rather than directly with the Longhorn UI application, `page` makes more sense for navigation.

c3y1huang: Brought this up in the meeting. It's @longhorn/doc 's decision.

PR #1071, opened 2025-03-12
doc: add exporting backing image details

ChanYiLin:
ref: https://github.com/longhorn/longhorn/issues/10093

### DIFF
--- a/content/docs/1.7.0/advanced-resources/backing-image/backing-image.md
+++ b/content/docs/1.7.0/advanced-resources/backing-image/backing-image.md
@@ -19,6 +19,10 @@ You can prepare a backing image using four different kinds of data sources.
 4. Restore a backing image from the backupstore, For more information, see [Backing Image Backup](../backing-image-backup).
 5. Clone a backing image.
 
+#### Volume exporting
+
+A backing image serves as the initial snapshot in the snapshot chain of a Longhorn volume. When you export a volume with an associated backing image, Longhorn merges that image with the delta changes, resulting in a new consolidated backing image.
+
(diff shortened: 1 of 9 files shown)


content/docs/1.7.0/advanced-resources/backing-image/backing-image.md
```diff
@@ -19,6 +19,9 @@ You can prepare a backing image using four different kinds of data sources.
 4. Restore a backing image from the backupstore, For more information, see [Backing Image Backup](../backing-image-backup).
 5. Clone a backing image.
 
+#### How Exporting Volume Works
+A backing image serves as the initial snapshot in the snapshot chain of a Longhorn volume. When export a backing image from an existing Longhorn volume, the volume may already have an associated backing image. If a volume with a backing image is exported, the process commits the original backing image and merges it with the delta changes, resulting in a new consolidated backing image.
```
jillian-maroket, line 23:
```suggestion
#### Volume exporting

A backing image serves as the initial snapshot in the snapshot chain of a Longhorn volume. When you export a volume with an associated backing image, Longhorn merges that image with the delta changes, resulting in a new consolidated backing image.
```

jillian-maroket (commented):
Review done. I don't think this is the best location for the info because the parent section is "Parameters during creation".

content/docs/1.7.0/advanced-resources/backing-image/backing-image.md
ChanYiLin: fixed.

jillian-maroket (approved):

PR #1088, opened 2025-04-24
Update export-from-replica.md

radwaanmi:
https://github.com/longhorn/longhorn/issues/10787 (wasn't able to open issue on longhorn/website for some reason - so closed it myself)
Issue # 

This PR is for editing the example of the RKE2/K3s example manifest to reflect rke2 default cni with required "hostNetwork: true" and keeping consistent with message and recommendation of mounting the volume with "readOnly: true" added to the manifest under volumeMounts section.

Wasn't able to open the issue on longhorn/website; not sure of why. closed issue inside of longhorn/longhorn since it wasn't related to proper repo, but referenced here since I did attempt an open of an issue.

Please review and correct any additions here if they are inadvisable, if linting isn't proper, verbiage is poor, etc as this is my first PR with SUSE.

### DIFF
--- a/content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
+++ b/content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
@@ -87,6 +87,7 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
      namespace: longhorn-system
    spec:
      hostPID: true
+     hostNetwork: true
      containers:
      - name: engine
        image: longhornio/longhorn-engine:v<current-version>
@@ -97,10 +98,13 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
        volumeMounts:
        - name: dev
          mountPath: /host/dev
+         readOnly: true
        - name: proc
          mountPath: /host/proc
+         readOnly: true
        - name: data
          mountPath: /volume
+         readOnly: true
      volumes:
      - name: dev
        hostPath:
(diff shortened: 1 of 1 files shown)


content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
```diff
@@ -87,6 +87,7 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
      namespace: longhorn-system
    spec:
      hostPID: true
+     hostNetwork: true
```
derekbit, line 90: Why set hostNetwork to `true`?

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
```diff
@@ -97,10 +98,13 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
        volumeMounts:
        - name: dev
          mountPath: /host/dev
+         readOnly: true
```
derekbit, line 101: longhorn needs to create a device. /host/dev shouldn't be read-only.

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
```diff
@@ -97,10 +98,13 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
...
        - name: proc
          mountPath: /host/proc
+         readOnly: true
```
derekbit, line 104: longhorn needs to create engine and replica processes. /host/proc shouldn't be read-only.

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
```diff
@@ -97,10 +98,13 @@ If the whole Kubernetes cluster or Longhorn system goes offline, the following s
...
        - name: data
          mountPath: /volume
+         readOnly: true
```
derekbit, line 107: Longhorn currently needs to update the volume metadata files while opening a replica directory. It shouldn't be read-only.

radwaanmi: Is there anyway to make it so whatever user spun up the container for the volume to only have read-only access or will that require using recursive read-only mounts? https://kubernetes.io/docs/concepts/storage/volumes/#recursive-read-only-mounts

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
radwaanmi: this is understandable it can be removed

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
radwaanmi: this is understandable it can be removed

content/docs/1.9.0/advanced-resources/data-recovery/export-from-replica.md
radwaanmi: Matt Mattox suggested this after he tested in his lab and found cases where this was needed. I believe it was needed for canal.

PR #1091, opened 2025-04-27
docs(1.9.0): update bakcup target creation notes and update table

vnwnv:
…tent

The `BackupTargetSecret` was removed in UI. There is no such field `Settings/General/BackupTargetSecret`.

Update the doc and add additional notes about create S3 compatible object storage section.

Update the topics table automatically.


vnwnv:
close with typo commit message

PR #1113, opened 2025-05-20
docs: update "Upgrade Path Enforcement and Downgrade Prevention"

derekbit:

Issue longhorn/longhorn#10945

### DIFF
--- a/content/docs/1.6.0/deploy/upgrade/_index.md
+++ b/content/docs/1.6.0/deploy/upgrade/_index.md
@@ -11,7 +11,7 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 # Upgrade Path Enforcement and Downgrade Prevention
 
-Starting with v1.5.0, Longhorn only allows upgrades from supported versions. When you attempt to upgrade from an unsupported version, the operation automatically fails but you can revert to the previously installed version without any service interruption or downtime.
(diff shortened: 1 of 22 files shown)


content/docs/1.6.0/deploy/upgrade/_index.md
```diff
@@ -11,7 +11,7 @@ There are no deprecated or incompatible changes introduced in v{{< current-versi
 
 # Upgrade Path Enforcement and Downgrade Prevention
 
-Starting with v1.5.0, Longhorn only allows upgrades from supported versions. When you attempt to upgrade from an unsupported version, the operation automatically fails but you can revert to the previously installed version without any service interruption or downtime.
+Starting with v1.5.0, Longhorn only allows upgrades from from one minor version to the next. For example, upgrading from 1.5.x to 1.6.x is allowed, but skipping minor versions (e.g., from 1.4.x to 1.6.x) is not supported. When you attempt to upgrade from an unsupported version or skip a minor version, the operation automatically fails. However, you can revert to the previously installed version without any service interruption or downtime.
```
sushant-suse, line 14:
```suggestion
Starting with v1.5.0, Longhorn only supports upgrades from one minor version to the next. For example, upgrading from 1.5.x to 1.6.x is supported, but skipping versions (e.g., from 1.4.x to 1.6.x) is not. If you attempt to upgrade from an unsupported version or skip a minor version, the operation will fail automatically. However, you can revert to the previously installed version without service interruption or downtime.
```

sushant-suse (approved):

PR #1140, opened 2025-06-24
docs: add Warning to Python Client

derekbit:

Issue Longhorn/longhorn#4284

### DIFF
--- a/content/docs/1.10.0/references/longhorn-client-python.md
+++ b/content/docs/1.10.0/references/longhorn-client-python.md
@@ -5,7 +5,7 @@ weight: 2
 
 > **Warning**
 >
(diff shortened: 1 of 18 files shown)


content/docs/1.10.0/references/longhorn-client-python.md
```diff
@@ -3,6 +3,10 @@ title: Python Client
 weight: 2
 ---
 
+> **Warning**
+>
+> The Longhorn Python client is not officially supported. It may change at any time without notice and without guarantee of backward compatibility.
```
sushant-suse, line 8:
```suggestion
> The Longhorn Python client is *not officially supported*. It may change at any time without notice and does **not** guarantee backward compatibility.
```

sushant-suse (changes requested):
The change was made for one version only but needs replicating to the other versions.
Thanks.

PR #1152, opened 2025-07-07
docs: add "Longhorn VolumeAttachment"

derekbit:

Issue Longhorn/longhorn#11142

### DIFF
--- a/content/docs/1.10.0/advanced-resources/volumeattachment.md
+++ b/content/docs/1.10.0/advanced-resources/volumeattachment.md
@@ -1,0 +1,246 @@
+---
+title: Longhorn VolumeAttachment
+weight: 1
+---
+
+**Table of Contents**
+- [Kubernetes and Longhorn `VolumeAttachment`](#kubernetes-and-longhorn-volumeattachment)
+- [Longhorn VolumeAttachment](#longhorn-volumeattachment)
+  - [VolumeAttachment CR](#volumeattachment-cr)
+  - [Understanding Attachment Tickets](#understanding-attachment-tickets)
+- [The CSI Attachment and Detachment Workflow](#the-csi-attachment-and-detachment-workflow)
+  - [Core Components in the Workflow](#core-components-in-the-workflow)
+  - [The CSI Volume Attachment Flow](#the-csi-volume-attachment-flow)
+  - [The CSI Volume Detachment Flow](#the-csi-volume-detachment-flow)
+  - [Summary of the Workflow](#summary-of-the-workflow)
+- [Troubleshooting Volume Attachment Issues](#troubleshooting-volume-attachment-issues)
+  - [Volume is Stuck in `Attaching` or `Detaching` State](#volume-is-stuck-in-attaching-or-detaching-state)
+    - [Possible Causes](#possible-causes)
+    - [Resolution Steps](#resolution-steps)
+  - [Case Study](#case-study)
+    - [Case 1: Failure to Attach Volume Due to Unexpected `longhorn-ui` Attachment Ticket](#case-1-failure-to-attach-volume-due-to-unexpected-longhorn-ui-attachment-ticket)
+    - [Case 2: Volume Fails to Attach to New Node Due to Backup Job Stuck in Pending State](#case-2-volume-fails-to-attach-to-new-node-due-to-backup-job-stuck-in-pending-state)
+
+---
+
+This document provides a detailed overview of Longhorn's `VolumeAttachment` custom resource, how it integrates with Kubernetes' native `VolumeAttachment`, its operational flow, and common troubleshooting scenarios.
+
+## Kubernetes and Longhorn `VolumeAttachment`
+
+To understand how volume attachments work in Longhorn, it is important to distinguish between the Kubernetes `VolumeAttachment` and Longhorn's custom `VolumeAttachment`:
+
+- **Kubernetes `VolumeAttachment`**: It is a native Kubernetes API resource that is part of the Container Storage Interface (CSI) specification. Its primary role is to signal to a CSI driver that a volume should be attached to a specific node.
+
+- **Longhorn `VolumeAttachment`**: It is a Custom Resource (CR) defined by Longhorn, with the full name `volumeattachment.longhorn.io`. This internal Longhorn resource is used by the Longhorn Manager to track and manage all attachment requests for a volume.
+
+## Longhorn VolumeAttachment
+
+### VolumeAttachment CR
+
+To retrieve a Longhorn `VolumeAttachment`, use the following command:
+
+```bash
+kubectl -n longhorn-system get volumeattachment.longhorn.io <volume-name> -o yaml
+```
+
+Example output:
+
+```bash
+apiVersion: v1
+...
+  spec:
+    attachmentTickets:
+      csi-f0471a334f0b249f964cd1dec461a5eb94c8d268cbbc904c1a8e9a37e2045208:
+        generation: 0
+        id: csi-f0471a334f0b249f964cd1dec461a5eb94c8d268cbbc904c1a8e9a37e2045208
+        nodeID: rancher60-master
+        parameters:
+          disableFrontend: "false"
+          lastAttachedBy: ""
+        type: csi-attacher
+    volume: pvc-b26e9514-aafd-46e0-b70c-4e3f187c7977
+  status:
+    attachmentTicketStatuses:
+      csi-f0471a334f0b249f964cd1dec461a5eb94c8d268cbbc904c1a8e9a37e2045208:
+        conditions:
+        - lastProbeTime: ""
+          lastTransitionTime: "2025-07-05T09:17:27Z"
+          message: ""
+          reason: ""
+          status: "True"
+          type: Satisfied
+        generation: 0
+        id: csi-f0471a334f0b249f964cd1dec461a5eb94c8d268cbbc904c1a8e9a37e2045208
+        satisfied: true
+...
+```
+
+- `spec.attachmentTickets`: A map containing all active attachment requests (also known as **tickets**). Each ticket includes:
+  - `id`: A unique identifier for each attachment ticket.
+  - `nodeID`: The ID of the node where the volume should be attached.
+  - `parameters`: Optional parameters for the attachment, such as `disableFrontend` and `lastAttachedBy`.
+  - `type`: The attacher type, indicating the source of the attachment request.
+
+- `status.attachmentTicketStatuses`: A map containing the current status of each active attachment ticket or request. Each entry includes:
+  - `conditions`: The current condition(s) of the ticket, including whether the request is satisfied or not.
+  - `satisfied`: A boolean value indicating whether the attachment request has been fulfilled or not.
+  - `generation`: The generation number of the ticket, used to track updates.
+
+### Understanding Attachment Tickets
+
+The Longhorn `VolumeAttachment` custom resource (CR) is designed to manage attachment requests from various internal Longhorn system controllers. Each request is represented as an **attachment ticket** within the CR.
+
+All active tickets are stored in the `spec.attachmentTickets` map. The `type` field in each ticket (referred to as the **AttacherType**) identifies the source of the request. The common `AttacherType` values include:
+
+- `csi-attacher`: The most common type which handles standard attachment requests from the Kubernetes CSI plugin, typically when mounting a volume to a pod.
+- `longhorn-api`: Represents a manual attachment request initiated by a user, either through the Longhorn UI or the Longhorn API.
+- `snapshot-controller`: Used when attaching a volume to create or restore a snapshot.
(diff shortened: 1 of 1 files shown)


shuo-wu (dismissed):
In general LGTM.

Do we need to provide a full flow starting from pod/Kubernetes VA to Longhorn VA then to Volume and Engine/Replicas?

derekbit:
> In general LGTM.
> 
> Do we need to provide a full flow starting from pod/Kubernetes VA to Longhorn VA then to Volume and Engine/Replicas?

Great idea. I will add this part.

content/docs/1.10.0/advanced-resources/volumeattachment.md
```diff
@@ -0,0 +1,102 @@
...
+- `volume-eviction-controller`: Handles attachments during the process of evicting a replica from a node.
+- `bim-ds-controller`: Used by the Backing Image Data Source Controller when creating a volume from a backing image.
+
+## Troubleshooting Volume Attachment Issues
+
+This guide provides possible causes and solutions for troubleshooting volume attachment issues. Always review system logs and your environment for specific details, and proceed with caution to avoid disrupting running workloads.
+
+1. Issue: Volume stuck in "attaching" state
+   - Possible Cause: Stale or orphaned attachment tickets in the `spec.attachmentTickets` map.
+   - Suggested Action: Inspect the attachmentTickets and remove any unnecessary tickets.
```
c3y1huang, line 94: This might be oversimplified. How can a user determine whether a ticket is unnecessary? Also, what are the consequences if a user accidentally removes an active ticket?

derekbit: It might be difficult to provide too many details, as we've encountered various cases. Do you have an example we can refer to?

derekbit: > This might be oversimplified. How can a user determine whether a ticket is unnecessary? 

It might be difficult to provide too many details, as we've encountered various cases. Do you have an example we can refer to? 

> user accidentally removes an active ticket?

The consequence of removing an active ticket is resulting in the volume detachment and interrupt the process like backup. This is explicitly explained in the purpose of the volumeattachment. Do we need to mention it again?

c3y1huang: > > This might be oversimplified. How can a user determine whether a ticket is unnecessary?
> 
> It might be difficult to provide too many details, as we've encountered various cases. Do you have an example we can refer to?
> 

Are there any past issues we can reference or link to for context?

> > user accidentally removes an active ticket?
> 
> The consequence of removing an active ticket is resulting in the volume detachment and interrupt the process like backup. This is explicitly explained in the purpose of the volumeattachment. Do we need to mention it again?

Users might not read the entire document. Should we add a direct link to that section for clarity?

c3y1huang: I’m just thinking that if we clarify this a bit more, we might be able to add troubleshooting support to the Longhorn CLI in the future.

c3y1huang: > Let me think if this makes sense because the doc is short enough...

That's fair. I'm also fine with the current.

derekbit: Added warning. Thank you.

content/docs/1.10.0/advanced-resources/volumeattachment.md
```diff
@@ -0,0 +1,227 @@
...
+- `snapshot-controller`: Handles attachments required to create or restore a volume snapshot.
+- `backup-controller`: Handles attachments required to perform a volume backup.
+- `volume-restore-controller`: Handles attachments required when restoring a volume from a backup.
+- `volume-clone-controller`: Handles attachments required when cloning a new volume from an existing one.
+- `share-manager-controller`: Specifically for ReadWriteMany (RWX) volumes, this controller manages the attachment of the backend volume to the share-manager pod.
+- `volume-expansion-controller`: Handles attachments required for online volume expansion.
+- `volume-rebuilding-controller`: Handles attachments required when rebuilding a degraded or missing replica.
+- `salvage-controller`: Handles attachments during the salvage process when Longhorn attempts to recover and re-attach a problematic volume.
+- `volume-eviction-controller`: Handles attachments during the process of evicting a replica from a node.
+- `bim-ds-controller`: Used by the Backing Image Data Source Controller when creating a volume from a backing image.
```
sushant-suse, line 86:
```suggestion
- `spec.attachmentTickets`: A map containing all active attachment requests (also known as **tickets**). Each ticket includes:
  - `id`: A unique identifier for each attachment ticket.
  - `nodeID`: The ID of the node where the volume should be attached.
  - `parameters`: Optional parameters for the attachment, such as `disableFrontend` and `lastAttachedBy`.
  - `type`: The attacher type, indicating the source of the attachment request.

- `status.attachmentTicketStatuses`: A map containing the current status of each active attachment ticket or request. Each entry includes:
  - `conditions`: The current condition(s) of the ticket, including whether the request is satisfied or not.
  - `satisfied`: A boolean value indicating whether the attachment request has been fulfilled or not.
  - `generation`: The generation number of the ticket, used to track updates.

### Understanding Attachment Tickets

The Longhorn `VolumeAttachment` custom resource (CR) is designed to manage attachment requests from various internal Longhorn system controllers. Each request is represented as an **attachment ticket** within the CR.

All active tickets are stored in the `spec.attachmentTickets` map. The `type` field in each ticket (referred to as the **AttacherType**) identifies the source of the request. The common `AttacherType` values include:

- `csi-attacher`: The most common type which handles standard attachment requests from the Kubernetes CSI plugin, typically when mounting a volume to a pod.
- `longhorn-api`: Represents a manual attachment request initiated by a user, either through the Longhorn UI or the Longhorn API.
- `snapshot-controller`: Used when attaching a volume to create or restore a snapshot.
- `backup-controller`: Used when attaching a volume to perform a backup.
- `volume-restore-controller`: Used when attaching a volume during a restore operation.
- `volume-clone-controller`: Used when attaching a volume for cloning from an existing volume.
- `share-manager-controller`: Manages backend volume attachments for ReadWriteMany (RWX) volumes by attaching them to the share-manager pod.
- `volume-expansion-controller`: Handles attachments needed for online volume expansion.
- `volume-rebuilding-controller`: Used when attaching a volume to rebuild a degraded or missing replica.
- `salvage-controller`: Used during the salvage process when Longhorn attempts to recover and reattach a problematic volume.
- `volume-eviction-controller`: Handles attachments involved in evicting a replica from a node.
- `bim-ds-controller`: Used by the Backing Image Data Source controller when creating a volume from a backing image.
```

content/docs/1.10.0/advanced-resources/volumeattachment.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Volume is Stuck in `Attaching` or `Detaching` State
+
+When a volume fails to complete the attach or detach operation for an extended period, the issue is often related to the attachment tickets within the `VolumeAttachment` CR.
+
+#### Possible Causes
+
+- **Stale or Orphaned Tickets**: An attachment ticket from a previous workload (e.g., a deleted pod or a completed backup job) was not properly cleaned up and remains in spec.attachmentTickets.
+
+- **Conflicting Tickets**: An existing ticket (e.g., an attachment request from the CSI attacher) prevents another operation (e.g., a manual detach) from proceeding.
```
sushant-suse, line 168:
```suggestion
Longhorn extends the native volume attachment mechanism of Kubernetes by introducing a custom `VolumeAttachment` CR. This design provides several advantages:

- **Decoupling and Abstraction**: The custom resource encapsulates complex attach or detach logic within Longhorn, reducing the responsibilities of the `longhorn-csi-plugin`.
- **Fine-Grained Control**: The attachment ticket system enables Longhorn to handle requests from multiple sources (for example, pods, snapshots, backups) while ensuring only one valid attachment per volume at any time.
- **Reliability and Traceability**: The CR gives clear visibility into attachment state and history of the volume, simplifying monitoring and troubleshooting.

In summary, the Kubernetes `VolumeAttachment` object initiates the attachment or detachment process, while Longhorn’s custom `VolumeAttachment` CR orchestrates and manages the actual operations internally.

## Troubleshooting Volume Attachment Issues

This section outlines common issues related to volume attachment and provides recommended resolution steps. Before making any changes, carefully inspect system logs and the relevant custom resources to avoid disrupting active workloads.

### Volume is Stuck in `Attaching` or `Detaching` State

When a volume remains in the `Attaching` or `Detaching` state for an extended period, the cause is often related to stale or conflicting attachment tickets in the Longhorn `VolumeAttachment` CR.

#### Possible Causes

- **Stale or Orphaned Tickets**: A ticket from a previous workload (for example, a deleted pod or completed backup job) was not properly removed and still exists under `spec.attachmentTickets`.

- **Conflicting Tickets**: An existing ticket (for example, from the CSI attacher) blocks a new request (for example, a manual detach or move to a different node).
```

content/docs/1.10.0/advanced-resources/volumeattachment.md
```diff
@@ -0,0 +1,227 @@
...
+  1. Inspect the Longhorn `VolumeAttachment` CR for the volume to check if any controller is holding onto the volume
+      ```bash
+      kubectl -n longhorn-system get volumeattachment.longhorn.io <volume-name> -o yaml
+      ```
+  2. If you see an attachment ticket from the backup controller, it means a backup job is locking the volume.
+  3. Do not delete the `backup-*` VolumeAttachment ticket directly. Longhorn will automatically recreate it. The correct approach is to **resolve the stuck backup job**.
+  4. Remove any `Backup` CRs with:
+       - `status.state = pending`
+       - `status.message` containing `Failed to get the Snapshot...`
+    This will release the volume and allow it to be reattached.
```
sushant-suse, line 227:
```suggestion
1. **Inspect the Longhorn `VolumeAttachment` CR**: Use the following command to view the attachment tickets:

    ```bash
    kubectl -n longhorn-system get volumeattachment.longhorn.io <volume-name> -o yaml
    ```

2. **Analyze Ticket Sources**: Look under `spec.attachmentTickets` and check the `type` field for each ticket to identify its source (for example, `csi-attacher`, `backup-controller`, etc.).

3. **Remove Invalid Tickets with Caution**: If you confirm a ticket is no longer needed (for example, its corresponding pod has been deleted), you may remove it by editing the CR.

> **Warning:**
> 
> Deleting an active ticket can cause serious disruptions. If you remove a ticket still required by a running workload, Longhorn interprets this as a detach request:
> 
> - The volume engine will stop on the node, causing the pod to lose storage access and encounter I/O errors, likely crashing the pod.
> - Kubernetes CSI will eventually detect the issue and re-attach the volume, recreating the ticket, but this causes downtime and may require manual pod restart.
> 
> Always verify that the workload related to the ticket is inactive before removing it.

4. **Verify the State**: After removing invalid tickets, Longhorn should be able to complete the attach or detach operation successfully.

### Case Study

#### Case 1: Failure to Attach Volume Due to Unexpected `longhorn-ui` Attachment Ticket

- **Issue**: Longhorn [#8339](https://github.com/longhorn/longhorn/issues/8339)  
- **Symptom**:  
  - Workloads using the affected volume remain stuck in `Pending` state.  
  - The Longhorn `VolumeAttachment` CR contains an unexpected ticket from `longhorn-ui`.  
- **Workaround**:  
  - Inspect the `VolumeAttachment` CR:  
    ```bash
    kubectl -n longhorn-system get volumeattachment.longhorn.io <volume-name> -o yaml
    ```  
  - If you find a `longhorn-ui` attachment ticket, remove the entire ticket block from the CR.

#### Case 2: Volume Fails to Attach to New Node Due to Backup Job Stuck in Pending State

- **Issue**: Longhorn [#10090](https://github.com/longhorn/longhorn/issues/10090)  
- **Symptom**:  
  - When a workload is rescheduled to a different node, the volume fails to follow.  
  - Backup jobs referencing non-existent snapshots remain stuck in `Pending` state, with `status.message` containing `failed to get the snapshot ... not found`.  
  - These stuck backup jobs hold onto the volume, blocking detach/reattach.  
- **Workaround**:  
  1. Check the Longhorn `VolumeAttachment` CR for any tickets locking the volume:  
      ```bash
      kubectl -n longhorn-system get volumeattachment.longhorn.io <volume-name> -o yaml
      ```  
  2. If you see a ticket from the backup controller, a backup job is locking the volume.  
  3. **Do not delete the `backup-*` attachment ticket directly**, as Longhorn will recreate it.  
  4. Instead, resolve the stuck backup job by removing any `Backup` CRs with:  
      - `status.state = pending`  
      - `status.message` containing `Failed to get the Snapshot...`  
  This releases the volume and allows it to be reattached.
```

sushant-suse (changes requested):
I have tried to make the content more user friendly as we discussed earlier. There are some suggestions for your consideration. Thanks.

sushant-suse (approved):

PR #1156, opened 2025-07-09
doc(replica): enable offline replica rebuilding by default

mantissahz:

Issue # longhorn/longhorn#11204


derekbit:
Moved to v1.11.0.

PR #1160, opened 2025-07-14
added warning about Document Volume Stability Risks

sushant-suse:

Issue [#11240](https://github.com/longhorn/longhorn/issues/11240)

It adds a warning regarding the impact of I/O latency, especially with HDDs, on Longhorn volume stability. The warning clarifies the critical role of latency in performance and prevent misunderstanding of IOPS/throughput metrics.

I have only updated the doc of v1.9.0.

N/A


content/docs/1.9.0/best-practices.md
```diff
@@ -42,6 +42,17 @@ We recommend the following setup for deploying Longhorn in production.
   - 500/250 max IOPS per volume (1 MiB I/O)
   - 500/250 max throughput per volume (MiB/s)
 
+> **Warning**: While Longhorn can function with HDDs (spinning disks) as storage, it is important to understand that **latency** plays a much more important role in volume stability than IOPS or throughput. This is especially true when multiple input-output intensive tasks are running, such as:
```
c3y1huang, line 45:
```suggestion
> **Warning**: While Longhorn can function with HDDs (spinning disks) as storage, it is important to understand that **latency** plays a much more important role in volume stability than IOPS or throughput. This is because HDDs are mechanical, relying on spinning platters and moving read/write heads to access data. This physical movement introduces inherent delays (seek time and rotational delay), leading to much higher latency compared to the SSDs or NVMe drives. This can directly cause instability when multiple input-output intensive tasks are running, such as:
```

innobead (approved):

derekbit (approved):

PR #1169, opened 2025-08-05
fix: storageclass example still uses the deprecated data engine fields

shuo-wu:

Issue https://github.com/longhorn/longhorn/issues/11409

The space trimming is applied automatically by the IDE. If it's not needed, I will update the PR


sushant-suse (dismissed):

derekbit (dismissed):

derekbit:
@shuo-wu Can you help resolve the conflicts? Thank you.

innobead (approved):

PR #1172, opened 2025-08-13
docs: update v1.10.0 settings

derekbit:

Issue Longhorn/longhorn#10829

### DIFF
--- a/content/docs/1.10.0/references/settings.md
+++ b/content/docs/1.10.0/references/settings.md
@@ -3,6 +3,7 @@ title: Settings
 weight: 1
 ---
 
+- [Value Format Types by Supported Data Engines](#value-format-types-by-supported-data-engines)
 - [Customizing Default Settings](#customizing-default-settings)
 - [General](#general)
   - [Node Drain Policy](#node-drain-policy)
@@ -42,6 +43,9 @@ weight: 1
   - [Long gRPC Timeout](#long-grpc-timeout)
   - [Offline Replica Rebuilding](#offline-replica-rebuilding)
   - [RWX Volume Fast Failover (Experimental)](#rwx-volume-fast-failover-experimental)
+  - [Data Engine Log Level](#data-engine-log-level)
+  - [Data Engine Log Flags](#data-engine-log-flags)
+  - [Replica Rebuilding Bandwidth Limit](#replica-rebuilding-bandwidth-limit)
 - [Snapshot](#snapshot)
   - [Snapshot Data Integrity](#snapshot-data-integrity)
   - [Immediate Snapshot Data Integrity Check After Creating a Snapshot](#immediate-snapshot-data-integrity-check-after-creating-a-snapshot)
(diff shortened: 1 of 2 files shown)


content/docs/1.10.0/references/settings.md
```diff
@@ -90,9 +94,31 @@ weight: 1
...
+  - Format: JSON object (e.g., `{"v1":"value1","v2":"value2"}`)
+  - Allows specifying different values for V1 and V2.
+
+- Data-engine-specific values for V1 Data Engine only
+  - Format: JSON object with `v1` key (e.g., `{"v1":"value1"}`)
+  - Only V1 Data Engine can be configured; V2 Data Engine is not affected by this setting.
+
+- Data-engine-specific values for V2 Data Engine only
+  - Format: JSON object with `v2` key (e.g., `{"v2":"value1"}`)
+  - Only V2 Data Engine can be configured; V1 Data Engine is not affected by this setting.
```
sushant-suse, line 121:
```suggestion
Each setting supports only one of the following formats, depending on its definition. The supported format determines which Data Engines can be configured and whether their values can differ.

- Single value for all supported Data Engines
  - Format: Non-JSON string (e.g., `1024`)
  - The value applies to all supported Data Engines and must be the same across them.
  - Data-engine-specific values are not allowed.
- Data-engine-specific values for V1 and V2 Data Engines
  - Format: JSON object (e.g., `{"v1": "value1", "v2": "value2"}`)
  - Allows specifying different values for V1 and V2 Data Engines.
- Data-engine-specific values for V1 Data Engine only
  - Format: JSON object with `v1` key only (e.g., `{"v1": "value1"}`)
  - Only the V1 Data Engine can be configured; the V2 Data Engine is not affected.
- Data-engine-specific values for V2 Data Engine only
  - Format: JSON object with `v2` key only (e.g., `{"v2": "value1"}`)
  - Only the V2 Data Engine can be configured; the V1 Data Engine is not affected.
```

content/docs/1.10.0/references/settings.md
```diff
@@ -511,20 +540,39 @@ This setting allows Longhorn to automatically rebuild replicas for detached volu
 
 Enable improved ReadWriteMany volume HA by shortening the time it takes to recover from a node failure.
 
+#### Data Engine Log Level
+
+> Default: `{"v2":"Notice"}`
+
+Applies only to the V2 Data Engine. Specifies the log level for the Storage Performance Development Kit (SPDK) target daemon. Supported values are: Error, Warning, Notice, Info, and Debug.
```
sushant-suse, line 547:
```suggestion
Applies only to the V2 Data Engine. Specifies the log level for the Storage Performance Development Kit (SPDK) target daemon. Supported values: `Error`, `Warning`, `Notice`, `Info`, and `Debug`.
```

content/docs/1.10.0/references/settings.md
```diff
@@ -972,11 +1012,11 @@ See [Trim Filesystem](../../nodes-and-volumes/volumes/trim-filesystem) for detai
 
 #### Guaranteed Instance Manager CPU
 
-> Default: `12`
+> Default: `{"v1":"12","v2":"12"}`
 
-Percentage of the total allocatable CPU resources on each node to be reserved for each instance manager pod when the V1 Data Engine is enabled. For example, Longhorn reserves 10% of the total allocatable CPU resources if you specify a value of 10. This setting is essential for maintaining engine and replica stability, especially during periods of high node workload.
+Percentage of the total allocatable CPU resources on each node to be reserved for each instance manager pod. For example, 10 means 10% of the total CPU on a node will be allocated to each instance manager pod on this node. This will help maintain engine and replica stability during high node workload.
```
sushant-suse, line 1017:
```suggestion
Percentage of the total allocatable CPU resources on each node to reserve for each instance manager pod. For example, a value of `10` means 10% of the total CPU on a node will be allocated to each instance manager pod on that node. This helps maintain engine and replica stability during periods of high node workload.
```

sushant-suse (changes requested):
Some suggestions for your consideration.
Thanks.

sushant-suse (approved):

PR #1184, opened 2025-08-29
feat: Add Google Analytics (GA4) tag to website

sushant-suse:

N/A

It adds the Google Analytics tag to the website to track user traffic and behaviour. This is essential for understanding how users interact with the site, which pages they visit most, and where they encounter issues. The data collected will help improve the website's user experience and content strategy.

Due to the custom and complex nature of the website's theme, the standard Hugo method of adding the Google Tag failed. So, I implemented a workaround by hardcoding the Google Tag script directly into the `layouts/partials/meta.html` file. This was necessary to ensure the tag is present on all pages, including the documentation and knowledge base sections, which were not correctly inheriting the base layout. I have also left the `google_tag.html` partial file in place as it is the part of the theme's structure, even though it is not currently being used for the Google Tag implementation.

#### Additional documentation or context:

- **Original Google Tag code**: The script used is from the official Google documentation for [`gtag.js`](https://developers.google.com/tag-platform/gtagjs#add_the_google_tag_to_your_website).
- **Troubleshooting**: The chosen solution was adopted after multiple attempts to use Hugo's built-in googleAnalytics variable and partials resulted in rendering errors due to the theme's specific templating context.
- **Configuration**: The googleAnalytics parameter was added to `config.toml`, but is not being used by the hardcoded solution. It is kept for future reference and for any potential future changes to the theme.


derekbit:
@sushant-suse Do we have any ticket for the improvement?

layouts/partials/meta.html
```diff
@@ -61,3 +61,12 @@
...
 {{ end }}
+
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-7FVLP2S3TJ"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
+  function gtag(){dataLayer.push(arguments);}
+  gtag('js', new Date());
+
+  gtag('config', 'G-7FVLP2S3TJ');
+</script>
```
houhoucoop, line 72: Redundant setting?

derekbit: cc @sushant-suse

PR #1193, opened 2025-09-08
docs(rebuilding): add resources insufficient section

mantissahz:

Issue # longhorn/longhorn#11270


content/docs/1.10.0/advanced-resources/rebuilding/offline-replica-rebuilding.md
```diff
@@ -29,6 +29,19 @@ Starting with v1.9.0, Longhorn supports offline replica rebuilding, allowing deg
 - When triggered, Longhorn attaches the volume without activating the frontend, rebuilds any missing replicas, and then detaches the volume upon completion.
 - This process can be interrupted if the workload scales up.
 
+## Resource Constraints
+
+Offline replica rebuilding will be cancelled or prevented from starting if there are insufficient resources available in the cluster.
+
+- All instance managers on schedulable nodes must be ready.
+- The scheduler verifies that there is an available disk with sufficient space on a schedulable node for rebuilding a replica.
```
derekbit, line 37: Can you reorganize the sentences? The two items don't have strong relationship with 
```
Offline replica rebuilding will be cancelled or prevented from starting if there are insufficient resources available in the cluster.
```

derekbit (commented):
Please reorganize the `Resource Constraints` and `Resources Insufficient`. It might be not easy for readers to understand.

@sushant-suse Could you go discuss with @mantissahz to improve the changes? Thanks.

content/docs/1.10.0/advanced-resources/rebuilding/offline-replica-rebuilding.md
sushant-suse, line 37: We can reframe it like:
```suggestion
Offline replica rebuilding will be canceled or prevented from starting if certain conditions are not met:

- All instance managers on schedulable nodes must be ready.
- The scheduler verifies that there is an available disk with sufficient space on a schedulable node for rebuilding a replica.

If these requirements are not met, Longhorn will consider the cluster to have insufficient resources.
```

content/docs/1.10.0/advanced-resources/rebuilding/offline-replica-rebuilding.md
```diff
@@ -29,6 +29,19 @@ Starting with v1.9.0, Longhorn supports offline replica rebuilding, allowing deg
...
+
+### Resources Insufficient
+
+- The volume remains in a **degraded state** and **detached** with fewer healthy replicas than desired.
+- During offline replica rebuilding starts, Longhorn will wait for an interval from as short as 30 seconds to the maximum `replica-replenishment-wait-interval` to detach the volume if resources are insufficient.
+- No automatic rebuilding starts until resources become available or are newly added.
```
sushant-suse, line 43: And this one can be reframed as:
```suggestion
If resources are insufficient:

- The volume remains in a degraded state and detached with fewer healthy replicas than desired.
- During offline replica rebuilding starts, Longhorn will wait for an interval from as short as 30 seconds to the maximum `replica-replenishment-wait-interval` to detach the volume.
- No automatic rebuilding starts until resources become available or are newly added.
```

sushant-suse (changes requested):

content/docs/1.10.0/advanced-resources/rebuilding/offline-replica-rebuilding.md
```diff
@@ -29,6 +29,21 @@ Starting with v1.9.0, Longhorn supports offline replica rebuilding, allowing deg
 - When triggered, Longhorn attaches the volume without activating the frontend, rebuilds any missing replicas, and then detaches the volume upon completion.
 - This process can be interrupted if the workload scales up.
 
+## Resource Constraints
```
derekbit, line 32: > ## Resource Constraints

The ticket is a bit unclear. When I read this section, I can't tell what message it’s supposed to deliver. Could you come up with a more suitable title?

content/docs/1.10.0/advanced-resources/rebuilding/offline-replica-rebuilding.md
```diff
@@ -29,6 +29,21 @@ Starting with v1.9.0, Longhorn supports offline replica rebuilding, allowing deg
...
+
+If these requirements are not met, Longhorn will consider the cluster to have insufficient resources:
+
+- All instance managers on schedulable nodes must be ready.
+- The scheduler verifies that there is an available disk with sufficient space on a schedulable node for rebuilding a replica.
```
derekbit, line 37: Again, the doc is for readers that don't know the context of the feature. Thus, it is better to organize the sentences as the steps

- Why is this section needed?
  - Problem
  - Proposed solution
  - Expected benefit
- How does it work?
  - High-level workflow
  - Key components involved
  - Example scenario
- Limitations and considerations
  - Known issues / trade-offs
  - Compatibility notes

You don't need to go into deep technical details, but you should clearly convey the feature and its importance to the users.

derekbit: cc @sushant-suse

PR #1216, opened 2025-10-06
Describe with proper accessMode

larssb:
### Issue that it fixes

incorrect accessMode mentioned in pvc-ownership-and-permission doc.

So that the correct accessmode is used


larssb:
Let me know if I need to update the same error in more versions of Longhorn. And here's the official accessMode description: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

PR #1338, opened 2026-03-30
docs: add KB article for NO_PROXY configuration

sushant-suse:

[Issue](https://github.com/longhorn/longhorn/issues/12522)

Adds a Knowledge Base article to resolve gRPC connectivity issues in proxied environments. It provides a guided configuration for the `NO_PROXY` environment variable to ensure internal Longhorn traffic (Manager-to-Instance Manager) bypasses external proxy servers, preventing `403 Forbidden` and `Unavailable` gRPC errors.

Verified this configuration on a live K3s cluster using a Squid proxy. Confirmed that without the Pod/Service CIDRs in `NO_PROXY`, the `longhorn-manager` failed to monitor engines and sync with the instance-manager. Applying the "Golden String" template successfully restored connectivity and volume health.

#### Additional documentation or context:

N/A

### DIFF
--- a/content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
+++ b/content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
@@ -1,0 +1,90 @@
+---
+title: "Troubleshooting: Prevent gRPC Proxying Between Internal Services by Setting NO_PROXY"
+authors:
+- "Sushant Gaurav"
+draft: false
+date: 2026-03-30
+versions:
+- All
+categories:
+- "troubleshooting"
+- "network"
+- "proxy"
+---
+
+## Applicable versions
+
+**Confirmed working with**:
+
+- Longhorn `v1.10.1` and later (contains gRPC code-level fixes)
+
+**Potentially applicable to**:
+
+- All Longhorn versions running in proxied Kubernetes environments (MicroK8s, RKE2, etc.)
+
+## Symptoms
+
+When Longhorn is installed in a cluster where a global proxy (for example, `HTTP_PROXY`, `HTTPS_PROXY`) is configured at the container runtime or OS level, you may encounter the following issues:
+
+- **Replica Failures**: Replica rebuilding, cloning, or restoration tasks fail with `rpc error: code = Unavailable desc = connection error`.
+- **Proxy Header Errors**: Logs in `longhorn-manager` showing `Invalid header received from client`. This typically happens when a proxy (like Privoxy or Squid) receives non-HTTP gRPC traffic and rejects it.
+- **Backup Failures**: Restoration from backup targets (NFS or S3) failing due to traffic being routed to an external gateway instead of internal endpoints.
+
+Example error log:
+```text
+replica pvc-xxxx-r-xxxx failed to get replica: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: failed to do connect handshake, response: \"HTTP/1.1 400 Invalid header received from client\""
+```
+
+## Reason
+
+Longhorn services communicate via gRPC. In environments with a global proxy, this internal traffic may be incorrectly routed through the proxy server.
+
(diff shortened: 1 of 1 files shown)


dariavladykina (dismissed):

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
```diff
@@ -0,0 +1,90 @@
...
+- "troubleshooting"
+- "network"
+- "proxy"
+---
+
+## Applicable versions
+
+**Confirmed working with**:
+
+- Longhorn `v1.10.1` and later (contains gRPC code-level fixes)
```
derekbit: The KB is applied for Longhorn version prior to v1.10.2 because we have fixed it since v1.10.2.
https://github.com/longhorn/longhorn/issues/12304

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
```diff
@@ -0,0 +1,90 @@
...
+    - name: NO_PROXY
+      value: "localhost,127.0.0.1,<SERVICE_CIDR>,<POD_CIDR>,.svc,.cluster.local"
+```
+
+### 3. Configure via Longhorn Settings
+
+For components managed dynamically by Longhorn (like `instance-manager` pods), ensure that the proxy settings are consistent:
+
+1. Navigate to Longhorn UI > Settings > General.
+2. If using a proxy for backups, ensure the Proxy Secret is configured.
```
derekbit: What is Proxy Secret?
cc @mantissahz

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
```diff
@@ -0,0 +1,90 @@
...
+Example error log:
+```text
+replica pvc-xxxx-r-xxxx failed to get replica: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: failed to do connect handshake, response: \"HTTP/1.1 400 Invalid header received from client\""
+```
+
+## Reason
+
+Longhorn services communicate via gRPC. In environments with a global proxy, this internal traffic may be incorrectly routed through the proxy server.
+
+While recent Longhorn versions include `grpc.WithNoProxy()` in the backend code, some traffic paths (such as inter-pod HTTP calls for backing image managers or health checks) and older versions of Longhorn still rely on environment variables. If the `NO_PROXY` variable is missing or incomplete, internal gRPC/HTTP traffic "escapes" the cluster, leading to connection resets by the proxy.
```
derekbit: We don't need the workaround after introducing `grpc.WithNoProxy()`
cc @c3y1huang

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
sushant-suse: I will update this line.

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
sushant-suse: Global setting found under Settings > General > Proxy Secret

content/kb/troubleshooting-prevent-gRPC-proxying-between-internal-services-by-setting-no-proxy.md
sushant-suse: Should I remove it entirely?

derekbit: Yes, and please check https://github.com/longhorn/longhorn/issues/12304 to refine the doc.

PR #1343, opened 2026-04-08
Fix relative links in nginx retirement notice

peyremorgan:
Updated links for Traefik Ingress and Gateway API in the Longhorn Ingress documentation to reflect the correct relative paths.

Issue https://github.com/longhorn/longhorn/issues/12904

The links in the deprecation notice in the docs page [Create an Ingress with Basic Authentication](https://longhorn.io/docs/1.12.0/deploy/accessing-the-ui/longhorn-ingress/) are broken. This PR fixes them.


peyremorgan:
Ignore this PR; @innobead's fix is better

PR #1346, opened 2026-04-16
docs(backuptarget): document enforced soft-mount options in nfsOptions

COLDTURNIP:

Issue longhorn/longhorn#12896

### DIFF
--- a/content/docs/1.12.0/snapshots-and-backups/backup-and-restore/set-backup-target.md
+++ b/content/docs/1.12.0/snapshots-and-backups/backup-and-restore/set-backup-target.md
@@ -7,7 +7,7 @@ A backup target is an endpoint used to access a backupstore. Backup targets can 
 
 {{< figure alt="the backup target UI page" src="/img/screenshots/backup-target/v1.10.0/page.png" >}}
 
-> **Note:**  
+> **Note:**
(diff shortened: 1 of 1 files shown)


content/docs/1.12.0/snapshots-and-backups/backup-and-restore/set-backup-target.md
```diff
@@ -420,19 +420,31 @@ Example:
...
 
 Example:
 
 ```text
-nfs://longhorn-test-nfs-svc.default:/opt/backupstore?nfsOptions=soft,timeo=330,retrans=3  
+nfs://longhorn-test-nfs-svc.default:/opt/backupstore?nfsOptions=rw,nolock
 ```
 
-Any mount options that you specify will replace, not add to, the default options.
+When `nfsOptions` is set, Longhorn merges the following options to ensure NFS operations time out instead of blocking indefinitely:
```
derekbit:
```suggestion
When `nfsOptions` is specified, Longhorn adjusts the options to ensure NFS operations fail instead of hanging indefinitely when the NFS service is unstable:

- `hard` is removed if present.
- `soft` is added if absent.
- `timeo` and `retry` are added with their default values (`timeo=300`, `retry=2`) only if you have not specified them. Custom values are preserved.
```

sushant-suse (approved):

derekbit (approved):

PR #1361, opened 2026-05-19
docs(v1.12.0): update "Best Practices"

derekbit:

Issue #

### DIFF
--- a/content/docs/1.12.0/best-practices.md
+++ b/content/docs/1.12.0/best-practices.md
@@ -5,7 +5,11 @@ weight: 4
 
 We recommend the following setup for deploying Longhorn in production.
 
+> **Important:** We recommend enabling only one data engine (V1 or V2) per cluster. When both data engines are enabled simultaneously, each node runs separate instance-manager pods for V1 and V2, each with its own guaranteed CPU reservation. This significantly increases CPU overhead per node (the V2 instance-manager pod alone consumes at least 1 dedicated CPU core, which is configurable, for the `spdk_tgt` process). If your workloads do not require both engines, disable the unused one to conserve resources. For more information about guaranteed CPU reservations and how CPU consumption differs between V1 and V2 upgrade flows, see [Guaranteed Instance Manager CPU](#guaranteed-instance-manager-cpu).
+
 - [Minimum Recommended Hardware](#minimum-recommended-hardware)
+  - [V1 Data Engine](#v1-data-engine)
+  - [V2 Data Engine](#v2-data-engine)
 - [Architecture](#architecture)
 - [Operating System](#operating-system)
 - [Kubernetes](#kubernetes)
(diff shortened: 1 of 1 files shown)


sushant-suse (dismissed):

content/docs/1.12.0/best-practices.md
```diff
@@ -5,7 +5,11 @@ weight: 4
 
 We recommend the following setup for deploying Longhorn in production.
 
+> **Important:** We recommend enabling only one data engine (V1 or V2) per cluster. When both data engines are enabled simultaneously, each node runs separate instance-manager pods for V1 and V2, each with its own guaranteed CPU reservation. This significantly increases CPU overhead per node (the V2 instance-manager pod alone consumes 1 dedicated CPU core for the `spdk_tgt` process). If your workloads do not require both engines, disable the unused one to conserve resources.
```
innobead:
```suggestion
> **Important:** We recommend enabling only one data engine (V1 or V2) per cluster. When both data engines are enabled simultaneously, each node runs separate instance-manager pods for V1 and V2, each with its own guaranteed CPU reservation. This significantly increases CPU overhead per node (the V2 instance-manager pod alone consumes at least 1 dedicated CPU core, which is configurable, for the `spdk_tgt` process). If your workloads do not require both engines, disable the unused one to conserve resources.
```

innobead: Suggest adding a reference link to CPU guarantee settings for clarity, since we avoid detailing everything about CPU usage here. For example, during the V1 data engine upgrade, the two instance manager pods use the same number of CPU core requests, whereas in V2, the upgrade is in-place with no additional CPU consumption beyond the instance manager pod.

content/docs/1.12.0/best-practices.md
```diff
@@ -220,7 +257,9 @@ Refer to [Guaranteed Instance Manager CPU](../references/settings/#guaranteed-in
...
 
-The `Guaranteed Instance Manager CPU for V2 Data Engine` setting allows you to reserve a specific number of millicpus on each node for each instance manager pod when the V2 Data Engine is enabled. By default, the Storage Performance Development Kit (SPDK) target daemon within each instance manager pod uses 1 CPU core. Configuring a minimum CPU usage value is essential for maintaining engine and replica stability, especially during periods of high node workload. The default value is 1250.
+The `Guaranteed Instance Manager CPU for V2 Data Engine` setting allows you to reserve a specific number of millicpus on each node for each instance manager pod when the V2 Data Engine is enabled. By default, the Storage Performance Development Kit (SPDK) target daemon (`spdk_tgt`) within each instance manager pod uses 1 CPU core. The `spdk_tgt` process is responsible for handling IO operations and requires intensive polling, consuming 100% of a dedicated CPU core to efficiently manage and process IO requests. Configuring a minimum CPU usage value is essential for maintaining engine and replica stability, especially during periods of high node workload. The default value is 1250.
```
innobead: For V2, are CPU cores required by the spdk_tgt part of this request, or is the request more for other resources needed in the Im pod?

https://longhorn.io/docs/1.11.2/references/settings/#data-engine-cpu-mask, still required regardless of this setting.

innobead (commented):
In general, LGTM. A few questions/suggestions.

content/docs/1.12.0/best-practices.md
derekbit: mainly for spdk_tgt, but im pod still consumes cup resource.

sushant-suse (approved):

PR #1378, opened 2026-06-01
docs(setting): update DataEngineCPUMask description and default to recommend 2+ cores

mantissahz:

Issue # longhorn/longhorn#13237


derekbit (approved):
Thanks @mantissahz

PR #1389, opened 2026-07-02
docs: nvme persistent device names

carterli0407-cell:
Issue [#13440](https://github.com/longhorn/longhorn/issues/13440)

Kernel-assigned NVMe device names (`/dev/nvmeXnY`) are not stable: they can change across reboots and differ between nodes for the same physical disk. This recently caused a test failure where the system disk enumerated as `nvme2n1` on
one AWS EC2 node instead of `nvme0n1`, breaking a hardcoded `/dev/nvme1n1` path.

This PR updates the "Multiple Disk Support" guide to explicitly warn users not to rely on kernel device names when adding V2 block-type disks, and to prefer stable identifiers instead:
- Use the disk's **BDF** for the `nvme` / `virtio-blk` drivers.
- Use a persistent symlink under `/dev/disk/by-id/` when using a device-path driver such as `aio`.


sushant-suse (dismissed):

derekbit (commented):
LGTM.

@carterli0407-cell 
- Could you update v1.12.0 and v1.12.1 docs? 
- The PR title should follow the conventional commit rules, e.g., `docs:....`. Please name it.
 
Thanks.

carterli0407-cell:
> LGTM.
> 
> @carterli0407-cell
> 
> * Could you update v1.12.0 and v1.12.1 docs?
> * The PR title should follow the conventional commit rules, e.g., `docs:....`. Please name it.
> 
> Thanks.

Updated. Thank you.

content/docs/1.12.0/nodes-and-volumes/nodes/multidisk.md
derekbit, line 187: @carterli0407-cell Please update the example

carterli0407-cell: Updated. Thank you.

derekbit (approved):

PR #1396, opened 2026-07-14
docs: use longhorn-crypto storage class to restore an encrypted volume

yangchiu:

Issue https://github.com/longhorn/longhorn-tests/pull/3550#discussion_r3575362640

use longhorn-crypto storage class to restore an encrypted volume


sushant-suse (approved):

derekbit (approved):
Good catch! Thanks @yangchiu 
cc @mantissahz

PR #1402, opened 2026-07-20
docs: use settings.longhorn.io in archives and KB

locker95:
https://github.com/longhorn/longhorn/issues/11690

#1198 updated the current Longhorn docs to use `settings.longhorn.io` (or `lhs`) instead of bare `settings`, so `kubectl` does not hit CRD name collisions.

This PR finishes the remaining spots that still used the short name:

- KB: `content/kb/troubleshooting-default-settings-do-not-persist.md`
- Archives: `content/docs/archives/1.2.3` … `1.9.3` `customizing-default-settings.md` (40 files)

Older archives that never documented `kubectl edit settings` are left alone. Current docs and 1.10+ archives were already fixed.

Sample change:

```diff
- kubectl edit settings <SETTING-NAME> -n longhorn-system
+ kubectl edit settings.longhorn.io <SETTING-NAME> -n longhorn-system
```

Plus the same collision note used in the current docs:

> To avoid collisions with other CRDs, do not use the simple `settings`. Instead, use `settings.longhorn.io` or `lhs`.

### DIFF
--- a/content/docs/archives/1.2.3/advanced-resources/deploy/customizing-default-settings.md
+++ b/content/docs/archives/1.2.3/advanced-resources/deploy/customizing-default-settings.md
@@ -168,7 +168,7 @@ It would make the setting persistent.
 If you prefer to use the command line to update the setting, you could use kubectl.
 But please be aware **this will bypass Longhorn backend validation**.
 ```shell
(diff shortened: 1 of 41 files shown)


derekbit (approved):
Thanks @locker95 for the update.

PR #1418, opened 2026-08-11
fix(docs): correct ublk kernel v6.17 limitation description in v1.12.0, v1.12.1, v1.13.0 docs

carterli0407-cell:
Issue [#11977](https://github.com/longhorn/longhorn/issues/11977)


derekbit (approved):
LGTM

PR #1434, opened 2026-08-25
docs: update note to remove dependency of v2 backing images

COLDTURNIP:

Issue longhorn/longhorn#13858

### DIFF
--- a/content/docs/1.12.0/important-notes/_index.md
+++ b/content/docs/1.12.0/important-notes/_index.md
@@ -49,12 +49,26 @@ V2 Backing Images are removed in Longhorn v{{< current-version >}}. Use the Cont
 
 **Migration required for existing V2 volumes with backing images:**
 
-If you have V2 volumes that were created from backing images, you must migrate them before upgrading to v{{< current-version >}}:
+> **Warning: Upgrade Failure Risk**
+> V2 volumes with backing image dependencies cannot be upgraded in-place. Attempting to upgrade without migration may result in volume attachment failures.
+>
+> Furthermore, a direct backup of a V2 volume does not include its V2 backing image data, meaning a restore would still require the original backing image. If the volume data is still needed, you **must** perform the migration below before upgrading. If the data is no longer needed, simply delete the V2 volume instead.
 
-1. **Backup and recreate** (recommended): Create a backup of the V2 volume, delete the original volume, then restore from backup. The restored volume will not have a backing image dependency.
-2. **Delete the volume**: If the data is not needed, delete the V2 volume directly.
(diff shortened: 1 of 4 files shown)


derekbit:
@COLDTURNIP Could you create a DOC ticket for the PR?
BTW, typo `dependency` in the commit and PR titles.

content/docs/1.12.0/important-notes/_index.md
```diff
@@ -49,10 +49,23 @@ V2 Backing Images are removed in Longhorn v{{< current-version >}}. Use the Cont
...
 
-If you have V2 volumes that were created from backing images, you must migrate them before upgrading to v{{< current-version >}}:
+A direct backup of a V2 volume does not include its V2 backing image data. Restoring that backup therefore still requires the original backing image. Before upgrading to v{{< current-version >}}, flatten each required V2 volume into a temporary V1 volume and back up the V1 volume:
 
-1. **Backup and recreate** (recommended): Create a backup of the V2 volume, delete the original volume, then restore from backup. The restored volume will not have a backing image dependency.
-2. **Delete the volume**: If the data is not needed, delete the V2 volume directly.
+1. **Stop writes to the source volume**: Detach the workload or otherwise ensure that no process can modify the V2 volume during the copy.
+2. **Create a temporary V1 volume**: Do not configure a backing image, and make the volume exactly the same size as the source V2 volume.
+3. **Copy every block**: Use an offline migration pod or maintenance environment that exposes both volumes as raw block devices on the same node. Copy the source device to the temporary V1 device, then verify the result.
```
sushant-suse:
```suggestion
> **Warning: Upgrade Failure Risk**
> V2 volumes with backing image dependencies cannot be upgraded in-place. Attempting to upgrade without migration may result in volume attachment failures. 
> 
> Furthermore, a direct backup of a V2 volume does not include its V2 backing image data, meaning a restore would still require the original backing image. If the volume data is still needed, you **must** perform the migration below before upgrading. If the data is no longer needed, simply delete the V2 volume instead.

Before upgrading to v{{< current-version >}}, flatten each required V2 volume into a temporary V1 volume and back up the V1 volume:

1. **Stop writes to the source volume:** Detach the workload or otherwise ensure that no process can modify the V2 volume during the copy.
2. **Create a temporary V1 volume:** Do not configure a backing image, and make the volume exactly the same size as the source V2 volume.
3. **Copy every block:** Use an offline migration pod or maintenance environment that exposes both volumes as raw block devices on the same node. Copy the source device to the temporary V1 device, then verify the result.
```

sushant-suse (changes requested):
Some suggestions for your consideration. Please apply to the other versions as well.
Thanks.

sushant-suse (approved):

PR #1438, opened 2026-09-02
docs(1.13.0): check-preflight extends check for iommu support

apoorvajagtap:
Issue https://github.com/longhorn/longhorn/issues/12076

Updates the following indicating how `check preflight` also reports about IOMMU support now:
- `Note` in interrup-mode-support
- list in `Check Prerequisites`

### DIFF
--- a/content/docs/1.13.0/deploy/install/_index.md
+++ b/content/docs/1.13.0/deploy/install/_index.md
@@ -196,8 +196,8 @@ The command used to install a NFSv4 client differs depending on the Linux distri
 
 You can also use the [Longhorn Command Line Tool](#longhorn-command-line-tool) to install `nfs-client` automatically.
 
-> **Notice:**  
-> These steps only verify that the kernel supports NFSv4, v4.1, or v4.2.  
+> **Notice:**
+> These steps only verify that the kernel supports NFSv4, v4.1, or v4.2.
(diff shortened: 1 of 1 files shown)


sushant-suse (dismissed):

content/docs/1.13.0/advanced-resources/v2-data-engine/interrupt-mode.md
```diff
@@ -40,6 +40,11 @@ Interrupt mode is particularly suitable for clusters with limited CPU resources
     If the command returns no output, IOMMU is not enabled.
 
     > **Note:** IOMMU support may not be exposed on virtualized instances. If unsure, consider using a bare-metal instance, or consult your cloud provider’s documentation or support team.
+    >
+    > Running the following also helps confirm IOMMU support on the instance:
```
derekbit: @c3y1huang Do we need IOMMU support while only using AIO disk? IIRC, we don't need this.

apoorvajagtap: Hello, I have removed this section, and just kept the `IOMMU support` intent. Let me know if this works. Thanks!

derekbit (approved):
LGTM. Thanks @apoorvajagtap

PR #1447, opened 2026-09-15
docs: document V2 live upgrade prerequisites, process, and recovery

davidcheng0922:

https://github.com/longhorn/longhorn/issues/9104


content/docs/1.13.0/deploy/upgrade/v2-live-upgrade.md
```diff
@@ -0,0 +1,68 @@
+---
+title: V2 Data Engine Live Upgrade
+weight: 5
+---
+
+V2 live upgrade upgrades instance managers one node at a time without detaching volumes. Longhorn temporarily relocates engines to other nodes, upgrades the instance manager, and restores the engines to their original nodes. Volumes remain attached, and workloads and host-side NVMe initiators stay on their original nodes. This process minimizes I/O interruption but does not guarantee uninterrupted I/O.
```
sushant-suse:
```suggestion
V2 live upgrade upgrades instance managers one node at a time without detaching volumes. Longhorn temporarily relocates engines to other nodes, upgrades the instance manager, and restores the engines to their original nodes. Volumes remain attached, and workloads and host-side NVMe initiators stay on their original nodes. 

> **Important**: This process minimizes I/O interruption, but it **does not guarantee uninterrupted I/O**.
```

content/docs/1.13.0/deploy/upgrade/v2-live-upgrade.md
```diff
@@ -0,0 +1,68 @@
...
+4. Longhorn upgrades one node at a time:
+
+   1. Temporarily relocate engines to eligible nodes.
+   2. Detach replica backends on the node being upgraded from their engines.
+   3. Update the instance manager container image and wait for it to become ready.
+   4. Restore engines to their original nodes.
+   5. Wait for affected volumes to become healthy as replicas are reused or rebuilt.
+
+   Nodes that only host replicas skip engine relocation and restoration. After a node completes its upgrade, Longhorn proceeds to the next node.
+5. Monitor progress:
```
sushant-suse:
```suggestion
1. Upgrade the Longhorn manager to the target version and wait for all manager pods to be running and ready.
2. In the Longhorn UI, go to **Settings** and configure the V2 upgrade settings before enabling automatic upgrade:
   - **Instance Manager Upgrade Timeout** (`instance-manager-upgrade-timeout`): Set the timeout in minutes. The default is 60 minutes.
3. Enable **Allow Instance Manager Automatic Upgrade** (`allow-instance-manager-automatic-upgrade`) for V2. This setting is disabled by default.
4. Longhorn upgrades one node at a time:
   1. Temporarily relocate engines to eligible nodes.
   2. Detach replica backends on the node being upgraded from their engines.
   3. Update the instance manager container image and wait for it to become ready.
   4. Restore engines to their original nodes.
   5. Wait for affected volumes to become healthy as replicas are reused or rebuilt.
   
   *(Note: Nodes that only host replicas skip engine relocation and restoration. After a node completes its upgrade, Longhorn proceeds to the next node).*

5. Monitor progress:
```

sushant-suse (changes requested):

content/docs/1.13.0/deploy/upgrade/v2-live-upgrade.md
davidcheng0922: Sorry, it is my description fault. I will fix it. Thank you

content/docs/1.13.0/deploy/upgrade/v2-live-upgrade.md
davidcheng0922: Fix, thank you

sushant-suse (dismissed):

content/docs/1.13.0/deploy/upgrade/_index.md
```diff
@@ -34,7 +34,7 @@ The following table outlines the supported upgrade paths.
 
 ## Manual Checks Before Upgrade
 Automated checks are only performed on some upgrade paths, and the pre-upgrade checker may not cover some scenarios.  Manual checks, performed using either kubectl or the UI, are recommended for these schenarios.  You can take mitigating actions or defer the upgrade until issues are addressed.
-- Ensure that all V2 Data Engine volumes are detached and the replicas are stopped.  The V2 Data Engine currently does not support live upgrades.
+- For V2 Data Engine volumes, check the [V2 live upgrade prerequisites](./v2-instance-upgrade/#prerequisites). If live upgrade is not supported for your configuration, detach all V2 volumes and ensure their replicas are stopped before upgrading.
```
derekbit: > If live upgrade is not supported for your configuration

When will it happen?

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,87 @@
+---
+title: V2 Data Engine Instance Manager Upgrade
+weight: 5
+---
+
+After upgrading the Longhorn manager, V2 instance managers must also be upgraded to the new version. Longhorn supports two upgrade methods:
```
derekbit, line 6: We need to mention and highlight the v2 live upgrade is not supported in 1-node cluster.

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,87 @@
...
+
+- **[Offline upgrade](#offline-upgrade):** Detach all V2 volumes, then upgrade instance managers. This is the simplest approach and works in all configurations.
+- **[Live upgrade](#live-upgrade):** Upgrade instance managers one node at a time without detaching volumes. This keeps volumes attached but requires specific prerequisites.
+
+## Offline Upgrade
+
+If live upgrade prerequisites are not met, or you prefer a simpler approach, use the offline method:
```
derekbit: Should mention the `Prerequisites` first, so the layout will be 

```
## Live Upgrade

### Prerequisites
....

## Offline Upgrade
```

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,87 @@
+---
+title: V2 Data Engine Instance Manager Upgrade
+weight: 5
+---
+
+After upgrading the Longhorn manager, V2 instance managers must also be upgraded to the new version. Longhorn supports two upgrade methods:
+
+- **[Offline upgrade](#offline-upgrade):** Detach all V2 volumes, then upgrade instance managers. This is the simplest approach and works in all configurations.
+- **[Live upgrade](#live-upgrade):** Upgrade instance managers one node at a time without detaching volumes. This keeps volumes attached but requires specific prerequisites.
```
derekbit:
```suggestion
- **[Live upgrade](#live-upgrade):** Upgrade instance managers one node at a time without detaching volumes. This keeps volumes attached but requires specific prerequisites.
- **[Offline upgrade](#offline-upgrade):** Detach all V2 volumes, then upgrade instance managers. This is the simplest approach and works in all configurations.
```

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,87 @@
+---
+title: V2 Data Engine Instance Manager Upgrade
+weight: 5
+---
+
+After upgrading the Longhorn manager, V2 instance managers must also be upgraded to the new version. Longhorn supports two upgrade methods:
+
+- **[Offline upgrade](#offline-upgrade):** Detach all V2 volumes, then upgrade instance managers. This is the simplest approach and works in all configurations.
```
derekbit: > Detach all V2 volumes, then upgrade instance managers. ...


Does user need to do anything to upgrade instance managers?

content/docs/1.13.0/deploy/upgrade/_index.md
davidcheng0922: I meant cases where the live upgrade prerequisites are not met, such as using the ublk frontend or having replicas on only one node. I'll reword this to make that explicit.

I will change the word to `prerequisites`.

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: No prob, I'll add it in the live upgrade section

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: ok, thankk you

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: ok, thank you

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: No. Once there are no running instances on an instance manager, controller automatically recreates the im pod with the new version.

sushant-suse (dismissed):

davidcheng0922:
I notice https://github.com/longhorn/website/pull/1451 already modified the `### Upgrade Behavior` part, I will remove my change here.

cc @derekbit cc @mantissahz

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,94 @@
...
+After upgrading the Longhorn manager, V2 instance managers must also be upgraded to the new version. Longhorn supports two upgrade methods:
+
+- **[Live upgrade](#live-upgrade):** Upgrade instance managers one node at a time without detaching volumes. This keeps volumes attached but requires specific prerequisites.
+- **[Offline upgrade](#offline-upgrade):** Detach all V2 volumes, then upgrade instance managers. This is the simplest approach and works in all configurations.
+
+## Live Upgrade
+
+### Prerequisites
+
+> **Warning:** V2 live upgrade is not supported in single-node clusters. Use an [offline upgrade](#offline-upgrade) instead.
```
derekbit:
```suggestion
> **Warning:** V2 live upgrade is not supported for single-node clusters. Use [offline upgrade](#offline-upgrade) instead.
```

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,94 @@
...
+
+- Upgrade from Longhorn `v1.12.2` or later.
+- Use the V2 Data Engine with the NVMe/TCP frontend. Live upgrade does not support ublk or sharded volumes.
+- Ensure V2 volumes are healthy and have healthy replicas on at least two different nodes. Each affected volume must have an available RW replica outside the node being upgraded.
+- Ensure nodes that temporarily host engines have a running V2 instance manager and a schedulable block disk.
+- Reserve capacity for replica rebuilding. Back up volumes and pre-download the target instance manager image on all V2 nodes before upgrading.
```
derekbit: How much capacity user should reserve? Could we give an estimation?

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,94 @@
...
+
+### Upgrade Considerations
```
derekbit, line 26: > pre-download the target instance manager image

Can we help pre-pull the image?

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
derekbit: > Back up volumes

Add it to Warning

e.g.
> 
> **Warning:**
> - V2 live upgrade is not supported in single-node clusters. Use an [offline upgrade](#offline-upgrade) instead.
>
> - Back up your V2 volumes before performing a live upgrade to protect against unexpected failures during the upgrade.

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
```diff
@@ -0,0 +1,94 @@
...
+
+Live upgrade upgrades instance managers one node at a time without detaching volumes. Longhorn temporarily relocates engines to other nodes, upgrades the instance manager, and restores the engines to their original nodes. Volumes remain attached, and workloads and host-side NVMe initiators stay on their original nodes. Live upgrade keeps I/O available.
```
derekbit:
```suggestion
Live upgrade upgrades instance managers one node at a time without detaching volumes. Longhorn temporarily moves the V2 Engine CRs (NVMe/TCP targets) to other nodes, upgrades the instance manager, and then moves the Engine CRs back to their original nodes. Throughout the process, volumes remain attached, while workloads and EngineFrontend CRs (host-side NVMe/TCP initiators) stay on their original nodes, allowing I/O to continue without interruption.
```

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: It depends on the volume size.

If a replica cannot be reused—for example, because the replenishment wait interval has expired or the replica is faulty—Longhorn needs capacity to create a new replica for rebuilding.

- If sufficient capacity is available, Longhorn creates a new replica and starts rebuilding.
- Otherwise, the volume waits until capacity is available, for example after orphaned replica data is deleted. This is controlled by the `orphan-resource-auto-deletion` setting.

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: Sure, but I’m not sure it is worth doing. Normally, replica rebuilding takes much longer than pulling the image, so pre-pulling may not significantly reduce the overall upgrade time. Our `instance-manager` image is around 500MB.

It is also not required for the upgrade. I think it would be better to move this statement from prerequisites to considerations.

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: Fixed, thank you

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: Fixed, thank you

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: Fixed, thank you

content/docs/1.13.0/deploy/upgrade/v2-instance-upgrade.md
davidcheng0922: One more concern is that our solution would likely deploy a DaemonSet to pre-pull the image. 

If all nodes pull the image at the same time, it could create a burst of network traffic and put pressure on the registry. I am not sure whether this would affect workloads in large clusters.

To avoid this, we would need to pre-pull the image gradually, for example node by node. That requires creating pods for specific nodes and managing their lifecycle and progress, which adds more complexity. Therefore, I am not sure whether introducing this mechanism is worthwhile here.

derekbit: I see. Another question, how can user manually pull the image?

davidcheng0922: IIUC, after k8s 1.24, using `crictl`

It looks like
```
sudo crictl pull longhornio/longhorn-instance-manager:v1.13.x-head
...
Image is up to date for sha256:11a8aef68938b0215526b283899ea58a852cd7ce6ba578c58644a70dcffd0a9c
```

I add this command in the document for reference.

derekbit: Might be not a good user experience because some distros don't have crictl.
Could you create an IMPROVEMENT ticket for this issue? We can figure out it according to feedback from user.

davidcheng0922: Sure, here is the ticket: https://github.com/longhorn/longhorn/issues/14046

derekbit (approved):

